PR #106, opened 2025-01-12
fix: listen_notify_local feature to avoid pulling in eventsource_client

audunhalland:
Hi, I'm evaluating this project and from the moment I found this on github I thought the idea behind it is quite awesome.

I'm only using the embedded mode, and enabling the `listen_notify` feature pulls in a problematic dependency: `eventsource_client`. It uses outdated versions of the hyper stack and I want to avoid the resulting bloat. This dependency is only used for the remote feature. Ideally, I think remote should be opt-in (perhaps even a separate crate?) because I think the value of this project is the embedded DB architecture.

However, this PR attempts to fix the dep issue now without breaking backwards-compatibility. It introduces a new feature `listen_notify_local`, which is the `listen_notify` feature just without the remote support. Let's see if it passes CI and if the general direction is agreeable.

### DIFF
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -48,10 +48,13 @@ full = [
     "webpki-roots",
 ]
 listen_notify = [
-    "dep:cryptr",
     "dep:eventsource-client",
+    "listen_notify_local"
+]
+listen_notify_local = [
+    "dep:cryptr",
     "dep:futures-util",
-    "cache",
+    "cache"
 ]
 s3 = ["dep:cryptr", "backup"]
 server = [
--- a/hiqlite/src/app_state.rs
+++ b/hiqlite/src/app_state.rs
@@ -13,7 +13,7 @@ use crate::client::stream::ClientStreamReq;
 use crate::dashboard::DashboardState;
 #[cfg(feature = "dlock")]
 use crate::store::state_machine::memory::dlock_handler::LockRequest;
-#[cfg(feature = "listen_notify")]
+#[cfg(feature = "listen_notify_local")]
 use crate::store::state_machine::memory::notify_handler::NotifyRequest;
 #[cfg(feature = "sqlite")]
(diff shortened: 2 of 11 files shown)


sebadob:
So, I took a look and yes we can do it like that. I guess in the long run, it probably makes sense to even split out a `listen_notify_remote` to further optimize the code, or maybe actually reverse the logic with the next bigger release, because in 99% of cases the feature is probably only used with an embedded replica instead of a remote node.

We also need to have a look at `fastwebsockets` + `axum-0.8` and when the direction with this is clear, it would make sense to migrate the remote listen / notify to a Websocket connection to get rid of this dependency completely.

But the checks are not passing with this PR yet. I do not have set up any github actions yet, so you would need to execute it locally.

Basically, you need to make sure that

```
just verify
```

is happy.

You may also need to update the `just clippy` recipe and add a sole `listen_notify_local` case. Additionally, there is a known issue with `clippy` checks in the current `openraft` release, so you probably need to change `cargo clippy -- -D warnings` to `cargo clippy` in `just check`.

audunhalland:
Will fix the PR. Curious about `fastwebsockets`, does it have any actual advantages over `tokio-tungstenite`?

sebadob:
> Curious about fastwebsockets, does it have any actual advantages over tokio-tungstenite?

It is more low level, but it has been significantly faster when I compared different crates a while ago.
On the other hand, `axum-0.8` improved the WebSocket performance and memory allocation and it may not be that big of a deal anymore to use the integrated handler directly. I have not done any testing with it.

sebadob:
I just pushed a [PR](https://github.com/denoland/fastwebsockets/pull/97) to upgrade `fastwebsockets` to `axum-0.8`. Was a super simple fix. So as soon as it (hopefully) gets merged and published, we can bump axum for hiqlite as well, and then eventually migrate the remote listener to websockets to drop the `eventsource-client` completely.

PR #117, opened 2025-02-06
feat: support more signing schemes in no-verify mode

audunhalland:
I have to use a Helm-generated certificate (https://helm.sh/docs/chart_template_guide/function_list/#genselfsignedcert) and it only outputs sha256WithRSAEncryption certs. After some hours of debugging I found out where the problem was.

This change extends the accepted algorithms to match the same set that `reqwest` supports: https://github.com/seanmonstar/reqwest/blob/37074368012ce42e61e5649c2fffcf8c8a979e1e/src/tls.rs#L642.


sebadob:
Even though in no verify mode, at least the 2 sha1 modes should be removed. They should never ever be used, not even with no verify.

audunhalland:
Done, removed both instances of sha 1

Issue #118, opened 2025-02-17
feature request: Refer to previous statement's RETURNING output in subsequent transaction statement


audunhalland:
Let's say I make a transaction consisting of two statements:

```
1: INSERT INTO foo (field) VALUES ('foo') RETURNING id;
2: INSERT INTO bar (field, foo_id) VALUES ('bar', $1)
```

I want the `$1` parameter of the second statement to somehow reference what the first statement returned. Since hiqlite state-machine writes have to be fully serialized+atomic there exists no way to execute arbitrary code between the statements in a transaction.

Suppose there was a `hiqlite::Param` variant that represented a _variable_, yet to be resolved to a value.  This variant would only be valid inside transactions. Suppose this was called `Param::StmtOutput(stmt_index, column_name)`. This would only work with statements that return a single row, which is probably fine for now (`INSERT` statements). The transaction execution loop would need to change into temporarily caching all the returned rows, so that the `StmtOutput` parameter can actually be resolved into a real value.

This is only an idea because I'm unsure how to continue with DB-native/autoincrement "surrogate key" design coupled with transactions. Up until now I've used UUIDs generated in the application, but they do take up a lot of space and I need to do a lot of crossreferencing.

audunhalland:
Thank you for the long comment, and I see what you mean, I can try to do it in pure SQL. But I don't consider it very "developer friendly". SQL statements will become potentially verbose and harder to understand, also potentially taking up more space in the Raft log(?).

So I tried to implement the idea in #119. I think doing it that way fits better with the mental model developers have of the power of SQL transactions: Somewhat tighter integration of application code and SQL. It could also be the case that the solution I came up with is too narrow and only solving a very specific problem, there might be use-cases for something _related_ that I just can't see, because I haven't been faced with that problem yet!

sebadob:
Implemented with +119

PR #119, opened 2025-02-18
feat: Param::StmtOutput

audunhalland:
A proposed implementation of the idea in #118 . 

I tried to do it with as little overhead as possible if the feature is not in use: 

Each executed statement pushes an `ExecutedStatement` onto a new `Vec`. That data structure contains the input SQL (`Cow<'static, str>`). That string is now kept around longer than it was before. The data structure also contains a vector representing the returned row. If the statement returned nothing, that will be an empty `Vec`, which does not heap-allocate.

If the feature _is_ in use, you may look up a previous row value by column name. Getting the column index from the column name is done lazily by trying to re-use cached statements.

It's technically a breaking change because it adds a new variant to `Param` and the enum isn't `#[non_exhaustive]`.

The naming of various things is up for discussion.

### DIFF
--- a/hiqlite/src/store/state_machine/sqlite/writer.rs
+++ b/hiqlite/src/store/state_machine/sqlite/writer.rs
@@ -5,13 +5,15 @@ use crate::store::state_machine::sqlite::state_machine;
 use crate::store::state_machine::sqlite::state_machine::{
     Params, StateMachineData, StateMachineSqlite, StoredSnapshot,
 };
+use crate::store::state_machine::sqlite::transaction_env::{ExecutedStatement, TransactionEnv, TransactionParamContext};
 use crate::{AppliedMigration, Error, Node, NodeId};
 use chrono::Utc;
 use flume::RecvError;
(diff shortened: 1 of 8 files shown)


hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -282,7 +284,7 @@ CREATE TABLE IF NOT EXISTS _metadata
                         q.tx.send(res).expect("oneshot tx to never be dropped");
                     }
 
-                    Query::Transaction(req) => {
+                    Query::Transaction(mut req) => {
```
audunhalland, line 287: This is a leftover from an experiment, it doesn't need `mut`

audunhalland:
The latest push improves two things:

1. Optionally refer to column by index, which completely avoids the re-prepare. I suppose it'll be most common to return only one column from txn statements.
2. Better user-facing way to create a variable parameters: `StmtIndex(n).column(col)` which converts into `Param`.

sebadob:
I like that a lot! Thank you very much.

PR #120, opened 2025-02-19
refactor: eliminate txn overhead when not using param variables

audunhalland:
After #119 was merged I had this idea on how to completely eliminate the memory usage overhead when not using statement outputs.

We now already branch on `column_count > 0`, it's clear that it's only those kinds of "observable" statements that need to be cached. So I changed the data structure from `Vec` to `HashTable<usize, ObservableStatement>`. That also means that the `Cow<'static, str>` that holds the input SQL does not need to have its lifetime extended when it's not returning any columns!

### DIFF
--- a/hiqlite/src/store/state_machine/sqlite/transaction_env.rs
+++ b/hiqlite/src/store/state_machine/sqlite/transaction_env.rs
@@ -4,44 +4,20 @@ use rusqlite::{types::Value, Transaction};
 
 use crate::Error;
 
-/// A previously executed statement
-pub struct ExecutedStatement {
-    /// The executed SQL
-    pub sql: Cow<'static, str>,
-    /// The first returned row of the statement.
-    /// This will be empty if the statement returned no rows.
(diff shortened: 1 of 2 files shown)


audunhalland:
The thinking is "don't pay for what you don't use". I'm thinking that in a transaction consisting of `n` statements, only a fraction of the statements will realistically return any columns. A majority will likely be "leaf statements", like in the example that inserts two parents and three children. So I think it makes sense to pay the "cost" of a dynamically growing HashTable (likely negligible in the big picture) only when using outputs and referencing. The memory usage and reallocations of the HashTable will be made up for by the SQL strings that can be discarded earlier (my understanding is that these are `Cow::Owned` on hiqlite nodes where the transaction did not originate)

I've not done any testing or perf measurement, it's only based on intuition of cost. If the HashTable is considered expensive because of the default hashing algorithm, there's always [fnv](https://docs.rs/fnv/latest/fnv/) which I tend to use for numeric keys when I'm in control of the key space.

PR #289, opened 2026-02-02
Even cache only data_dir must exist

hardcodes:
Recently I could not start hiqlite in cache only mode and just received an OS error "not writable". The reason was that even in cache only mode `data_dir` must exist and be writable.

This pull request adds some error message to the fact and some documentation line to the top level file `hiqlite.toml`.

### DIFF
--- a/hiqlite.toml
+++ b/hiqlite.toml
@@ -41,6 +41,8 @@ listen_addr_api = "0.0.0.0"
 listen_addr_raft = "0.0.0.0"
 
 # The data dir hiqlite will store raft logs and state machine data in.
+# `data_dir` must exist and be writable, even if cache only is used!
+# See also `cache_storage_disk`.
 #
 # default: data
 # overwritten by: HQL_DATA_DIR
@@ -141,6 +143,7 @@ wal_size = 2097152
 # all your cached data will be gone.
 # In-memory only hugegly increases the throughput though, so it
 # depends on your needs, what you should prefer.
+# See also `data_dir`.
 #
 # default: true
 # overwritten by: HQL_CACHE_STORAGE_DISK
--- a/hiqlite/src/error.rs
+++ b/hiqlite/src/error.rs
@@ -88,6 +88,9 @@ pub enum Error {
     WAL(String),
     #[error("WebSocket: {0}")]
     WebSocket(String),
+    #[cfg(feature = "toml")]
+    #[error("Error: {0}")]
+    String(String),
 }
 
 impl Error {
@@ -158,6 +161,8 @@ impl IntoResponse for Error {
             Error::Unauthorized(_) => StatusCode::UNAUTHORIZED,
             Error::WAL(_) => StatusCode::INTERNAL_SERVER_ERROR,
             Error::WebSocket(_) => StatusCode::BAD_REQUEST,
+            #[cfg(feature = "toml")]
+            Error::String(_) => StatusCode::INTERNAL_SERVER_ERROR,
         };
 
         (status, Json(self)).into_response()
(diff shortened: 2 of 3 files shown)


bartekus:
Took a look at this against current `main`. The underlying issue is still present, and digging into it I think it confirms @sebadob's instinct that a rework is the cleaner path.

Where the disk requirement comes from in in-memory mode (`cache_storage_disk = false`, i.e. `in_memory_only = true`):

* `store/mod.rs` builds the memory state machine with `in_memory_only = !cache_storage_disk`.
* In `store/state_machine/memory/state_machine.rs`, `StateMachineMemory::new()` still calls `fs::create_dir_all(&path_snapshots)` and `set_path_access(...)` unconditionally (the `in_memory_only` branch above it only wipes the dir first), so `data_dir` must exist and be writable even in pure in-memory mode.
* The deeper reason is that raft snapshots are always disk-backed: `type SnapshotData = fs::File`, and `build_snapshot` / `begin_receiving_snapshot` / `install_snapshot` / `get_current_snapshot` all read and write `fs::File` under `path_snapshots`. The stored `in_memory_only` flag is only used to clear the dir at startup; it does not gate the snapshot I/O.

So in-memory mode keeps the cache data in memory but still persists raft snapshots to disk, which is why `data_dir` is still required.

Two directions from here:

1. This PR as-is (clearer error plus the doc note): useful as an interim, but it currently conflicts with `main` so it would need a rebase, and it keeps the disk requirement.
2. The rework hinted at above: when `cache_storage_disk = false`, hold snapshots in memory (for example switch `SnapshotData` to a `Cursor<Vec<u8>>`-style buffer behind the `in_memory_only` flag) and skip the `create_dir_all` / `set_path_access` calls. That makes in-memory-only genuinely disk-free. It is bounded but touches the four snapshot methods, so it is a real change rather than a one-liner.

Happy to help with either: rebase this one, or open a fresh PR for the in-memory-snapshot rework, whichever you all prefer.

sebadob:
> The deeper reason is that raft snapshots are always disk-backed

Exactly. Not doing this would mean double the memory usage. Snapshots are only (maybe) necessary when a new node joins the cluster. This snapshot is mandatory for consistency, but almost never needed.

This could be resolved, but let's say you have lots of data in your cache. You could then need 4GB of memory instead of 2GB (roughly). This is nothing anyone would want. Having a dir on disk available is the solution you always want. If you e.g. do not want to use a `StatefulSet` inside K8s, you could use a `Deployment` with an ephemeral Volume. This does not matter at all. Data can be gone after shutdown.

So, I'm not sure if this is even something desriable at all. But yes, it could be made a "feature" with an additional opt-in.

bartekus:
I dug into *why* `data_dir` is required in cache-only mode and wanted to share what I found, plus an alternative worth considering.

The requirement comes down to a single spot. Even with `cache_storage_disk = false`, the cache Raft log store is already fully in-memory (`LogStoreMemory`), but `StateMachineMemory` still does `create_dir_all({data_dir}/state_machine_cache/snapshots)` and writes its snapshots there. So a pure cache-only, in-memory node touches `data_dir` only for those snapshot files.

That also happens to contradict the `cache_storage_disk` docs, which say `false` keeps "the cache WAL **+ Snapshots** ... in memory": the WAL is in memory, but the snapshots were not.

So rather than requiring `data_dir` in this mode, we can make `cache_storage_disk = false` actually do what its docs promise and keep snapshots in memory too, which removes the `data_dir` requirement entirely for fully in-memory cache-only nodes. I opened #340 with that approach: `SnapshotData` becomes `Cursor<Vec<u8>>`, all `data_dir` access is gated on the in-memory flag, disk-backed behavior is unchanged, plus a unit test asserting nothing is written to disk.

I see the two as complementary rather than competing. When `data_dir` genuinely is used (disk-backed cache, or SQLite), the clearer error message proposed here is still the right thing. #340 just removes the requirement for the one case that motivated this issue. Happy to go whichever direction you prefer.

sebadob:
> That also happens to contradict the cache_storage_disk docs, which say false keeps "the cache WAL + Snapshots ... in memory": the WAL is in memory, but the snapshots were not.

Right, I guess we should update the documentation, mentioning that the snapshots will be stores on disk by default (as mentioned in the other PR). This could even be mostly ignored, because for in-memory nodes, it's basically just a temp file. Snapshots lived in-memory only when I created the whole thing, but I changed it later for way better efficiency. I just did not update the description.

sebadob:
superseded by #340

PR #317, opened 2026-04-14
fix: cookie header format for secure cookies

kinolaev:
Fixed the `Path` attribute for session cookies. Cookies with the `__Host-` prefix must include `Path=/` per the [Cookie Prefixes spec](https://datatracker.ietf.org/doc/html/draft-west-cookie-prefixes-05#section-3.2).


sebadob:
Thanks, I missed that missing `;` there. I usually expose via HTTP and do local port forwarding.

PR #331, opened 2026-05-24
Support learner-only startup mode

mrchypark:
## Summary

- add `learner_only` / `HQL_LEARNER_ONLY` configuration so a newly joining node can stay as a learner during startup reconciliation
- treat configured learner-only nodes as ready when they are learner members instead of requiring voter membership, while preserving the existing non-learner voter readiness checks
- document the config in generated/sample config files, README, and changelog
- add unit and cluster coverage for learner-only readiness and non-voter membership

## Validation

- PASS: `RUSTFLAGS="--cfg tokio_unstable" /Users/cypark/.cargo/bin/cargo +nightly test --workspace --features cache,counters,dlock,listen_notify,macros,toml --lib --bins`
- PASS: `RUSTFLAGS="--cfg tokio_unstable" /Users/cypark/.cargo/bin/cargo +nightly test --features cache,counters,dlock,listen_notify,macros,toml --test cluster learner_only_node_stays_non_voter_and_becomes_ready -- --nocapture`
- PASS: `RUSTFLAGS="--cfg tokio_unstable" /Users/cypark/.cargo/bin/cargo +nightly clippy --workspace --features cache,counters,dlock,listen_notify,macros,toml --lib --bins -- -D warnings`
- PASS: `RUSTFLAGS="--cfg tokio_unstable" /Users/cypark/.cargo/bin/cargo +nightly clippy -p hiqlite --features cache,counters,dlock,listen_notify,macros,toml --test cluster -- -D warnings -A clippy::needless_borrow`
- NOTE: `cargo fmt --check` and strict `clippy --tests` still report pre-existing issues outside this PR: repo-wide rustfmt import/order diffs, existing `hiqlite-wal` test clippy warnings, and existing cluster test needless-borrow warnings.
- NOTE: the full cluster test command still includes S3 restore scenarios and was not used as the final validation gate in this local environment.


sebadob:
Since this is still a Draft, I guess you are still working on it, right?

mrchypark:
This is ready for review now. I added learner-only readiness coverage, documented the new config, and updated the validation notes in the PR body.

hiqlite/src/network/api.rs
```diff
@@ -185,6 +162,47 @@ pub async fn ready(state: AppStateExt) -> Result<(), Error> {
...
+        warn!("not yet a ready member of the {raft_label} raft");
+        return Err(Error::Error(
+            format!("not yet a ready member of the {raft_label} raft").into(),
+        ));
+    }
+
+    Ok(())
+}
+
+fn is_ready_member_state(
```
sebadob: There is need to extract this logic into a new fn (and another one). Makes it harder to follow, and actually ineffeicient. E.g. we don't even need to extract `is_voter` or `is_member` when `state == ServerState::Shutdown`. Also, we can return early if `is_voter == true`.

I guess you did it to be able to test more easily below. To make testing `ensure_ready_member()` directly clean, we could only accept the `id` and `learner_only` and not the full `AppStateExt`. From the Metrics maybe only the `membership_config` and `state`, which should be possible to build pretty easily inside the tests. Then you would have 5 distinct args and can return early without building possibly unnecessasry data.

sebadob (changes requested):
Apart from this one thing about the ready check, looks good so far.

sebadob:
> NOTE: cargo fmt --check and strict clippy --tests still report pre-existing issues outside this PR: repo-wide rustfmt import/order diffs, existing hiqlite-wal test clippy warnings, and existing cluster test needless-borrow warnings.

Probably because of an updated Rust version in the meantime. I always clean them up and just follow the official Rust release.

> NOTE: the full cluster test command still includes S3 restore scenarios and was not used as the final validation gate in this local environment.

I have a full env running with multiple S3s, but I guess it's a good idea for the future to add a `just` recipe for starting a test container for that purpose. I will keep that in mind.

hiqlite/src/network/api.rs
mrchypark: Applied in 322a293. I changed ensure_ready_member() to take only the node id, learner_only flag, ServerState, and membership config, then return early for Shutdown and voter-ready cases before checking the learner-only path. The tests now build StoredMembership fixtures and exercise ensure_ready_member() directly. Verified with the targeted readiness unit tests, the learner-only cluster test, and clippy for the hiqlite lib target.

sebadob:
Thanks!

PR #333, opened 2026-06-04
Release 0.14.0-202606024: bump rusqlite 0.40, fs4 1.1, tikv-jemallocator 0.7

bartekus:
Bumps the SQLite stack and other lagging dependencies, and cuts the breaking `0.14.0-202606024` release. Updated per @sebadob's review.

## Dependency bumps (workspace)

| crate | before | after | notes |
|---|---|---|---|
| rusqlite | 0.39 | 0.40 | pulls libsqlite3-sys 0.37 -> 0.38 |
| serde_rusqlite | 0.42 | 0.43 | lockstep with rusqlite; pulls serde_core |
| fs4 | 0.13.1 | 1.1.0 | API change, see below |
| tikv-jemallocator | 0.6.0 | 0.7.0 | drop-in |

## Versions

All three crates bumped to `0.14.0-202606024` (breaking change). MSRV is `1.95` (raised by serde_rusqlite 0.43, which pulls serde_core). The inter-crate dependency requirements are left at "0.13" since 0.14 is unpublished; assuming they bump at publish time per the justfile flow.

## fs4 1.x change

fs4 1.x mirrors std's now-stable file locking (Rust 1.89):

* `fs4::fs_std::FileExt` becomes `fs4::FileExt` (trait moved to crate root)
* `try_lock_exclusive` becomes `try_lock`, returning `Result<(), TryLockError>` instead of `Result<bool>`

On MSRV 1.95, std's inherent `File::try_lock` shadows the trait method, so `hiqlite-wal/src/lockfile.rs` calls fs4 explicitly via `FileExt::try_lock(...)`. Since std now covers this natively, fs4 could alternatively be dropped here entirely; flagged on the thread for your call.

## Notes

* No `cargo fmt --check`: the tree is not default-rustfmt-clean, so a gate would fail until a one-time reformat.
* No source changes for rusqlite 0.40: hiqlite implements no custom VTabs (the 0.40 breaking changes are all in the VTab API), and every enabled feature still exists in 0.40.

## Validation (Rust 1.95)

* `cargo clippy -- -D warnings` (workspace) clean.
* `cargo check -p hiqlite --features full,jemalloc` clean.
* `hiqlite-wal` clippy clean with the fs4 fix.
* All example crates build; lockfiles regenerated.

### DIFF
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -4,7 +4,8 @@ version = "0.13.2"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
-rust-version = "1.88.0"
+# Bumped to 1.95 by serde_rusqlite 0.43 (required for rusqlite 0.40), which depends on serde_core.
+rust-version = "1.95.0"
 categories = ["database", "caching"]
 keywords = ["database", "sql", "sqlite", "raft", "cache"]
 description = "Hiqlite - highly-available, embeddable, raft-based SQLite + cache"
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -1,10 +1,9 @@
 [package]
 name = "hiqlite"
-version = "0.13.2"
+version = "0.14.0-202606024"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
-# Bumped to 1.95 by serde_rusqlite 0.43 (required for rusqlite 0.40), which depends on serde_core.
 rust-version = "1.95.0"
 categories = ["database", "caching"]
 keywords = ["database", "sql", "sqlite", "raft", "cache"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,7 @@ reqwest = { version = "0.13", default-features = false, features = [
     "json",
     "rustls-no-provider",
 ] }
-rusqlite = { version = "0.39", features = [
+rusqlite = { version = "0.40", features = [
     "backup",
     "bundled",
     "cache",
@@ -95,7 +95,7 @@ rustls = { version = "0.23.37", default-features = false, features = [
 ] }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.57"
-serde_rusqlite = "0.42"
+serde_rusqlite = "0.43"
 sha2 = { version = "0.11", features = [] }
 spow = { version = "0.6", features = ["server"] }
 thiserror = "2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ fastwebsockets = { version = "0.10.0", features = [
     "unstable-split",
 ] }
 flume = "0.12"
-fs4 = { version = "0.13.1", features = ["sync"] }
+fs4 = { version = "1.1.0", features = ["sync"] }
 futures-util = "0.3.30"
 getrandom = { version = "0.4", features = ["std"] }
 hex = "0.4.3"
@@ -100,7 +100,7 @@ sha2 = { version = "0.11", features = [] }
 spow = { version = "0.6", features = ["server"] }
 thiserror = "2"
 thread-priority = "3.0.0"
-tikv-jemallocator = "0.6.0"
+tikv-jemallocator = "0.7.0"
 time = "0.3.47"
 tokio = { version = "1.43.1", features = ["fs", "sync", "rt-multi-thread"] }
 tokio-rustls = { version = "0.26.4", features = ["ring"] }
--- a/examples/bench/Cargo.lock
+++ b/examples/bench/Cargo.lock
@@ -1409,7 +1409,7 @@ checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
 name = "hiqlite"
-version = "0.13.2"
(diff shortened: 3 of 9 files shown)


hiqlite/Cargo.toml
```diff
@@ -4,7 +4,8 @@ version = "0.13.2"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
-rust-version = "1.88.0"
+# Bumped to 1.95 by serde_rusqlite 0.43 (required for rusqlite 0.40), which depends on serde_core.
```
sebadob: That comment does not provide any useful information. We don't care why in comments, only in a PRs description, maybe. Here we only care about what the MSRV is.

sebadob (changes requested):
When bumping versions it makes sense to bump all of them that are behind, not just a few. Meaning we also need to bump:

- `fs4`
- `tikv-jemallocator`

Also, since this is a breaking change, please bump the main version in all crates to `0.14.0-202606024`.

bartekus:
Done, pushed.

* Dropped the comment above `rust-version`; only `rust-version = "1.95.0"` remains.
* Bumped the other lagging deps: `fs4` 0.13.1 to 1.1.0, `tikv-jemallocator` 0.6.0 to 0.7.0.
* Bumped all three crates to `0.14.0-202606024`.
* Regenerated the example lockfiles.

One thing worth flagging on the `fs4` bump. fs4 1.x mirrors std's file-locking API (stabilized in Rust 1.89), so it now overlaps with std. Two concrete effects in `hiqlite-wal/src/lockfile.rs`:

* `fs4::fs_std::FileExt` is gone; the trait moved to the crate root (`fs4::FileExt`).
* `try_lock_exclusive` was renamed to `try_lock` and now returns `Result<(), TryLockError>` instead of `Result<bool>`.

And because our MSRV is now 1.95, std's own inherent `File::try_lock` shadows the trait method, so `file.try_lock()` resolves to std rather than fs4. I disambiguated to fs4 explicitly with `FileExt::try_lock(file)` so we keep using fs4 as you asked.

That raises a question though: since the MSRV is now >= 1.89, std covers this lock natively, and `hiqlite-wal` is the only fs4 user (this one exclusive lock). Do you want to keep fs4 (current state), or switch to `std::fs::File::try_lock` and drop the dependency? Either is a one-line difference, your call.

Two smaller notes:

* I left the inter-crate requirements (`hiqlite` -> `hiqlite-wal`/`hiqlite-derive`) at "0.13", since 0.14 is not published yet and the committed manifests reference the registry versions. I assume those bump at publish time per your justfile flow; happy to bump them and switch to path deps if you'd rather the PR be self-consistent pre-publish.
* I used the version string exactly as given: `0.14.0-202606024`.

Verified on Rust 1.95: `cargo clippy -- -D warnings` (workspace) is clean, `cargo check -p hiqlite --features full,jemalloc` is clean, and all example crates build.

hiqlite/Cargo.toml
bartekus: My bad, sorry!

bartekus:
Sorry, got lazy and used GitHub branch rebase (result: Unverified commit 👎 ) instead of doing it properly locally then pushing (result: verified commit 👍 )

Before merging, please acknowledge my notes for further refinement.

sebadob:
> That raises a question though: since the MSRV is now >= 1.89, std covers this lock natively, and hiqlite-wal is the only fs4 user (this one exclusive lock). Do you want to keep fs4 (current state), or switch to std::fs::File::try_lock and drop the dependency? Either is a one-line difference, your call.

Back when I created `hiqlite-wal`, the file lock guarantees from `std` were not the same and weaker than the ones from `fs4`. If I remember correctly, even when locked, the `std` still allowed opening the file, or something like that. I would keep `fs4` for now, but open an issue to revisit the guarantees.

> I left the inter-crate requirements (hiqlite -> hiqlite-wal/hiqlite-derive) at "0.13", since 0.14 is not published yet and the committed manifests reference the registry versions. I assume those bump at publish time per your justfile flow; happy to bump them and switch to path deps if you'd rather the PR be self-consistent pre-publish.

They should be path-based by default. Otherwise, it would not be possible to do development or testing. Only during the release pipeline, they are changed to the released versions. But even then, this must be done in order. Otherwise it won't work and checks will fail. That's why there are both versions with one of them always out-commented. I just forgot to reverse the setting after the latest release.

> Verified on Rust 1.95: cargo clippy -- -D warnings (workspace) is clean, cargo check -p hiqlite --features full,jemalloc is clean, and all example crates build.

The big one is that `just test` is fine. But the latest `rusqlite` did not have breaking changes, since we dont use vtables.

sebadob (approved):
Looks good now. Thanks.

bartekus:
Opened #338 to track the `fs4` vs std lock-guarantee comparison (call site, the std stabilization details, and the cross-platform check). Agreed on keeping `fs4` until those guarantees are verified.

On the inter-crate deps: thanks for explaining the path-vs-version toggle. Since `main` currently has them on the version side ("0.13"), want me to open a quick PR flipping `hiqlite` -> `hiqlite-wal`/`hiqlite-derive` back to the path deps (with the commented version lines updated to `0.14.0-202606024` for the next release)? Happy to, just say the word.

And good catch on `rusqlite`: right, no vtable usage here, so 0.40 was non-breaking for us; the breaking part of the release was really the MSRV bump.

Issue #334, opened 2026-06-04
Add CI (GitHub Actions) to run `just verify` on push/PR


bartekus:
## Summary

The repository has no `.github/workflows` (only `FUNDING.yml`), so contributions are not validated automatically. Everything in `just verify` (clippy across feature combinations, `clippy-examples`, `minimal-versions check`, the `tests/cluster` suite, and `msrv-verify`) currently runs only locally on the maintainer's machine. This issue proposes adding a GitHub Actions CI workflow that mirrors `just verify`, so the same gates run on every push and pull request.

## Why this matters

* No automated guard on contributions: PRs (including Dependabot bumps) can change behavior, break a feature-flag combination, or fail `minimal-versions` resolution without anything catching it before merge.
* A concrete example surfaced in #333 (`rusqlite` 0.39 to 0.40): `serde_rusqlite 0.43` pulls in `serde_core` and requires Rust 1.95, which raises the crate's declared MSRV from 1.88 to 1.95. With no MSRV job, exactly this class of change can land unnoticed; an MSRV gate makes the floor explicit and enforced.
* The `justfile` already encodes the maintainer's definition of "everything is good". CI should run it so local and remote verification cannot diverge.

## Proposed implementation

A single `ci.yml` that drives the existing `justfile` as the source of truth, split into parallel jobs for clear signal:

| job | mirrors | toolchain |
|---|---|---|
| lint | `just clippy` + `just clippy-examples` | stable |
| check | `just check` (`cargo clippy -- -D warnings` + `cargo minimal-versions check`) | stable + nightly |
| test | `just test` (`--cfg tokio_unstable`, the `tests/cluster` suite) | nightly |
| msrv | `just msrv-verify` (per-crate `cargo msrv verify`) | declared `rust-version` |

Engineering choices:

* Drives `just`, so CI can never drift from local verification: one source of truth.
* All actions pinned by commit SHA; least-privilege `permissions: contents: read`; `Swatinem/rust-cache` for build caching; `concurrency` cancels superseded runs on the same ref.
* `just`, `cargo-hack`, `cargo-minimal-versions`, and `cargo-msrv` installed from prebuilt binaries via `taiki-e/install-action` (fast, no compile-from-source).

Deliberately out of scope for the first pass:

[...]

PR #335, opened 2026-06-04
Add GitHub Actions CI (runs `just verify`)

bartekus:
Implements #334. Reworked per @sebadob's review.

## What

Adds `.github/workflows/code_style.yaml`: a single job that runs inside `ghcr.io/sebadob/rauthy-builder:20260519` using only the first-party `actions/checkout`. No third-party actions or images. Steps run sequentially so the workspace compiles once, and draft PRs are skipped. Structure mirrors rauthy's `code_style.yaml`.

Steps:

* `cargo clippy -- -D warnings`
* `just clippy` (the curated feature matrix)
* `just clippy-examples`

## Scope and the image

I pulled and introspected `rauthy-builder:20260519`. It ships stable 1.95, `just`, clippy, rustfmt, `cc` and node, which fully covers the clippy/code-style checks above. It does not currently include a nightly toolchain, `cargo-msrv`, `cargo-minimal-versions`, or `cargo-hack`, so the heavier `just verify` steps cannot run in it as-is:

* `just test`: needs `cargo +nightly`
* `just check`: needs `cargo-minimal-versions` (plus nightly)
* `just msrv-verify`: needs `cargo-msrv`

Options for those are laid out in the review thread (lightest: keep code-style only; or a tool-free MSRV guard via `rustup toolchain install`; or extend the builder image). The MSRV guard is the #334 motivation, so I would suggest at least that one, but it is your call.

## Notes

* No `cargo fmt --check`: unlike rauthy, hiqlite's tree is not default-rustfmt-clean (import ordering), so a gate would fail until a one-time reformat. Happy to add both if wanted.
* `just clippy` is the one heavier step (it recompiles the feature combinations); trivial to drop to just `cargo clippy -- -D warnings` for rauthy-level leanness.
* Drafts are skipped by design, so marking this PR ready for review triggers the first run.

## Validation

* `cargo clippy -- -D warnings`, `just clippy`, and `just clippy-examples` all run clean locally (exit 0).
* `actionlint` passes on the workflow.
* Verified inside the actual `rauthy-builder:20260519` image that `cargo`, `clippy`, and `just` resolve and the recipes parse.

Resolves and closes issue #334


sebadob:
I definitely don't want to pull in any external actions or even images. That's exactly how tons of pipelines got hacked in the recent weeks.

You can make use of the rauthy-builder image. It contains `just` and should have all the tools we could need here as well:

```yaml
check:
    if: github.event.pull_request.draft == false
    name: Check Code Style
    runs-on: ubuntu-latest
    container:
      image: ghcr.io/sebadob/rauthy-builder:20260519
    steps:
```

Also, since Github runners are super slow, I would run multiple steps in a single job instead of multiple jobs, so we only need to compile once.

bartekus:
Thanks, that all makes sense. Pushed a rework:

* No third-party actions or images: one job inside `ghcr.io/sebadob/rauthy-builder:20260519`, only `actions/checkout@v4`, mirroring your `code_style.yaml`.
* Sequential steps in a single job, so we compile once. Drafts skipped, same `pull_request` types you use.
* Steps: `cargo clippy -- -D warnings`, then hiqlite's `just clippy` feature matrix and `just clippy-examples`.

One finding from pulling and introspecting `rauthy-builder:20260519`: it has stable 1.95 + `just` + clippy/rustfmt + cc + node, which covers the clippy checks fully. It does not currently ship a nightly toolchain, `cargo-msrv`, `cargo-minimal-versions`, or `cargo-hack`, so the heavier `just verify` parts cannot run in it as-is:

* `just test` uses `cargo +nightly` (the recipe pins nightly for doc auto-cfg under `--cfg tokio_unstable`)
* `just check` uses `cargo-minimal-versions` (plus nightly)
* `just msrv-verify` uses `cargo-msrv`

How would you like to handle those (lightest first)?

1. Keep CI to code-style only, like rauthy, and keep test/msrv local.
2. Add a tool-free MSRV guard using the image as-is: `rustup toolchain install $MSRV && cargo +$MSRV check` per crate. That covers the original motivation (#334) without needing `cargo-msrv`.
3. Extend a future `rauthy-builder` tag (or `cargo install` in a step) with nightly plus those cargo tools, so the full `just verify` runs in CI.

Two smaller things:

* I left `cargo fmt --check` out: unlike rauthy, hiqlite is not default-rustfmt-clean (import ordering), so it would fail until a one-time reformat. Glad to add the gate plus the reformat if you want it.
* `just clippy` (the feature matrix) is the one heavier step; happy to drop it to just `cargo clippy -- -D warnings` if you want it as lean as rauthy.

sebadob:
I just had a quick look. The `auto_doc` feature changed recently anyway. We don't need nightly anymore.

But while checking, I saw that the new check for left-over config values needs a fix. When we have a config that contains e.g. values for the dashboard like `password_dashboard`, but the `dashboard` feature is disabled, these values will never be read and are left-overs at the end, which then returns an error. I guess this check needs to have a whitelist for possible values that belong to features that may be disabled. The `just test` will only work currently when the `dashboard` feature is also enabled. But that would be another PR.

sebadob:
I just fixed the config left-over values check and remove the nightly requirement for running tests in #336.

bartekus:
Nice, #336 makes this simple. Since tests no longer need nightly and the left-over config check is fixed, the stable `rauthy-builder` image can run them directly. Pushed.

The single job now runs, in order, all on stable inside `rauthy-builder:20260519`:

1. `cargo clippy -- -D warnings`
2. `just clippy` (feature matrix)
3. `just clippy-examples`
4. `just test`

One job, one checkout, no third-party actions.

Still left out, since the image does not ship the tooling for them:

* `just check`'s `cargo minimal-versions check` (needs cargo-minimal-versions + cargo-hack)
* `just msrv-verify` (needs cargo-msrv)

If you want MSRV guarded in CI without adding tools, I can drop in `rustup toolchain install $MSRV && cargo +$MSRV check` per crate (rustup is already in the image). Otherwise happy to leave both local-only.

Two small questions:

* The job is still named "Check Code Style" but now also runs the test suite. Want it renamed (e.g. just "Check"), or kept as is?
* I trimmed the explanatory header comment per your note on #333, and kept one line on the `TERM` env (the `clear` at the top of each recipe fails without it on a runner).

Validation: the clippy matrix and examples are clean locally, and I am running `just test` on stable 1.95 to confirm the suite is green without nightly.

sebadob:
> One job, one checkout, no third-party actions.

That's perfect. I like to keep things simple.

> Still left out, since the image does not ship the tooling for them:
>
>    just check's cargo minimal-versions check (needs cargo-minimal-versions + cargo-hack)
>    just msrv-verify (needs cargo-msrv)

That's totally fine. They run during the release pipeline, and a new release needs a PR anyway.

> The job is still named "Check Code Style" but now also runs the test suite. Want it renamed (e.g. just "Check"), or kept as is?

Don't really care about the name. It was just a copy & paste from Rauthy. Just `Check` would be fine as well.

Running `just test` in CI would be nice, but I am not sure if it's worth it. It should work just fine, but they take quite some time.

sebadob:
Let's merge it an see how long it takes for the `just test` to finish.

Thanks

bartekus:
> Let's merge it an see how long it takes for the `just test` to finish.
> 
> Thank

You are most welcome, and back at you, thank you for your (ongoing) work on Hiqlite good sir!

sebadob:
Just pushed #337 to see how the CI behaves.

PR #337, opened 2026-06-04
Fix: revert internal deps to path-based

sebadob:
Changes dependencies to resolved locally again during development. 

The main purpose though is to see how the CI works out.

### DIFF
--- a/.github/workflows/code_style.yaml
+++ b/.github/workflows/code_style.yaml
@@ -1,4 +1,4 @@
-name: Code Style
+name: Check
 
 on:
   pull_request:
@@ -16,7 +16,7 @@ env:
 jobs:
   check:
     if: github.event.pull_request.draft == false
-    name: Check Code Style
+    name: Check
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/sebadob/rauthy-builder:20260519
@@ -29,4 +29,5 @@ jobs:
       - name: Clippy (examples)
         run: just clippy-examples
       - name: Tests
-        run: just test
+        # S3 tests cannot be done inside GitHub CI
+        run: just test-no-s3
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -81,10 +81,10 @@ webpki-roots = [
 ]
 
 [dependencies]
-hiqlite-derive = { version = "0.13", optional = true }
-#hiqlite-derive = { path = "../hiqlite-derive", optional = true }
-hiqlite-wal = { version = "0.13" }
-#hiqlite-wal = { path = "../hiqlite-wal" }
+#hiqlite-derive = { version = "0.13", optional = true }
+hiqlite-derive = { path = "../hiqlite-derive", optional = true }
+#hiqlite-wal = { version = "0.13" }
+hiqlite-wal = { path = "../hiqlite-wal" }
 
 argon2 = { workspace = true, optional = true }
 axum.workspace = true
--- a/hiqlite/src/error.rs
+++ b/hiqlite/src/error.rs
@@ -118,6 +118,7 @@ impl Error {
         None
     }
 
+    #[cfg(feature = "toml")]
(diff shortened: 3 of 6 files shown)


bartekus:
Diagnosed the `Tests` failure. The clippy steps and every cluster test pass; it dies only at the **S3 backup-restore** step:

```
panicked at hiqlite/tests/cluster/main.rs:70:13:
S3: No `S3Config` given, cannot restore backup
```

Root cause: `hiqlite/tests/cluster/main.rs:50` loads S3 credentials from a local `config` file via `dotenvy::from_filename("../config")`. That file is not in the repo, so CI has no `HQL_S3_URL`, and the S3 restore (`start_test_cluster_with_backup(false)`) has no `S3Config`. The local-file restore right after it does not use S3 and is fine.

Fix: run the S3 restore block only when S3 is configured, and pass the running cluster through when it is skipped, so the file-restore and self-heal tests still run:

```rust
    // BEGIN: test S3 backup restore
    let (client_1, client_2, client_3) = if env::var("HQL_S3_URL").is_ok() {
        log("Change current database as preparation for backup restore tests");
        backup::test_backup_restore_prerequisites(&client_1).await?;
        log("Current database has been changed");

        log("Shutting down nodes");
        join_all([
            client_1.shutdown(),
            client_2.shutdown(),
            client_3.shutdown(),
        ])
        .await;

        // just give the s3 upload task in the background some time to finish
        time::sleep(Duration::from_millis(1000)).await;

        log("Trying to start the cluster again after shutdown with restore from S3 backup");
        let (client_1, client_2, client_3) =
            backup_restore::start_test_cluster_with_backup(false).await?;
        log("Cluster has been started again");

        start::wait_for_healthy_cluster(&client_1, &client_2, &client_3).await?;
        log("Cluster is healthy again");
        (client_1, client_2, client_3)
    } else {
        log("Skipping S3 backup restore test (HQL_S3_URL not set)");
        (client_1, client_2, client_3)
    };
    // END: test S3 backup restore
```

`env` is already imported in `main.rs`. Locally (with your `config` / `HQL_S3_URL` present) the S3 path runs exactly as before; in CI it is skipped. Compile-checked against current `main`.

Want me to open a PR with this, or will you fold it into #337?

sebadob:
I will fix this tomorrow. The other `clippy`s pass, but they shouldn't. They show issues, so we should add the `-D` option here as well to make sure everything is clean. 

I also have some memory optimizations on my TODO, and will then push a new release.

bartekus:
Agreed, good catch. Since CI just drives the justfile, the cleanest place to add it is in the `clippy` and `clippy-examples` recipes themselves: append `-- -D warnings` to each `cargo clippy` call (same as `check` already does). That way local `just verify` and CI stay in sync and fail on the exact same lints, with no CI-only divergence.

Sounds great on the memory optimizations and the release. Happy to send a PR for the `-D warnings` recipe change (and/or the S3 test skip) if it saves you time, otherwise I will leave it with you. Thanks for merging the bumps and getting CI in.

sebadob:
17m for a pipeline is quite a lot, but at least it checks everything. I think that's okay. Especially for a project like this where you can easily mess up something, since the internal logic is pretty complex.

bartekus:
Agreed, and I think 17m is the right call here. For a distributed SQLite + Raft layer where a subtle ordering or persistence bug can corrupt state across a whole cluster, a pipeline that genuinely exercises everything is worth far more than a fast but shallow one: a green check actually meaning green is exactly what you want when the internal logic is this involved.

It's also not locked in. If it ever becomes a bottleneck there's plenty of headroom to trim without losing coverage: caching the cargo registry and target dirs between runs, splitting the independent test groups into parallel jobs, and gating the heavier S3/backup paths behind a path filter or label so they only run when relevant. So I wouldn't optimize it preemptively; let it prove its value first, and we can shave minutes off later if it starts to hurt.

Net: thoroughness over speed is the correct default for this project. Nice work getting CI in.

PR #340, opened 2026-06-05
Don't require data_dir for in-memory cache-only nodes

bartekus:
## Summary

Make a pure cache-only node run fully in-memory when `cache_storage_disk = false`, so that `data_dir` is no longer required to exist or be writable. This is an opt-in alternative to requiring `data_dir` in cache-only mode (see #289).

## Background

#289 documents that even in cache-only mode `data_dir` must exist and be writable, and improves the error when it can't be created. While tracing *why* that requirement exists, it comes down to a single place: the in-memory cache state machine always persists its snapshots to disk, even when the cache WAL is kept in memory.

For a pure cache-only node with `cache_storage_disk = false`:

- the cache Raft log store is already fully in-memory (`LogStoreMemory`), but
- `StateMachineMemory` still calls `create_dir_all({data_dir}/state_machine_cache/snapshots)` and writes snapshot files there.

That also contradicts the `cache_storage_disk` doc, which promises that `false` keeps "the cache WAL **+ Snapshots** ... in memory". The WAL was in memory; the snapshots were not.

## Change

- `TypeConfigKV::SnapshotData`: `tokio::fs::File` -> `Cursor<Vec<u8>>`. This decouples the Raft snapshot transport type from on-disk persistence. The snapshot is already fully serialized into a `Vec<u8>` before being written, and received snapshots were already read back into a `Vec<u8>` before applying, so the memory profile is unchanged. This is the same `SnapshotData` type openraft's own mem-store examples use, and the chunked `InstallSnapshotRequest` transport is unaffected.
- `StateMachineMemory`: when `in_memory_only` (i.e. `cache_storage_disk = false`), keep the latest snapshot in memory and never touch `data_dir` in `new` / `build_snapshot` / `begin_receiving_snapshot` / `install_snapshot` / `get_current_snapshot`.
- Disk-backed behavior (`cache_storage_disk = true`) is unchanged: snapshots are still written to and restored from `data_dir`.
- Docs (`hiqlite.toml`, `NodeConfig` doc comments) now explain that `data_dir` is not required for a pure cache-only, in-memory node.
- Added a unit test asserting an in-memory cache node creates nothing on disk and can still build and retrieve a snapshot.

## Relationship to #289

Complementary, not competing: when `data_dir` genuinely *is* used (disk-backed cache, or SQLite), the clearer error message from #289 is still the right thing. This PR removes the requirement entirely for the fully in-memory cache-only case, which is the scenario that motivated the issue.

## Testing

- `cargo clippy --no-default-features --features cache` and `--features sqlite,cache`: clean.
- `cargo clippy -- -D warnings` (default features): clean.
- New unit test passes. The existing cluster test suite already runs with `cache_storage_disk = false`, so it exercises the in-memory snapshot path end-to-end, including snapshot transfer during self-healing.

### DIFF
--- a/hiqlite.toml
+++ b/hiqlite.toml
@@ -41,6 +41,10 @@ listen_addr_api = "0.0.0.0"
 listen_addr_raft = "0.0.0.0"
 
 # The data dir hiqlite will store raft logs and state machine data in.
+# It must exist and be writable, unless you run a pure cache-only node (no
+# SQLite) with `cache_storage_disk = false`: that mode keeps everything
+# in-memory and never touches `data_dir`.
+# See also `cache_storage_disk`.
 #
 # default: data
 # overwritten by: HQL_DATA_DIR
@@ -141,6 +145,9 @@ wal_size = 2097152
 # all your cached data will be gone.
 # In-memory only hugegly increases the throughput though, so it
 # depends on your needs, what you should prefer.
+# In a pure cache-only deployment (no SQLite), `false` also means `data_dir`
+# is never used and does not need to exist.
+# See also `data_dir`.
 #
 # default: true
 # overwritten by: HQL_CACHE_STORAGE_DISK
--- a/hiqlite.toml
+++ b/hiqlite.toml
@@ -41,8 +41,9 @@ listen_addr_api = "0.0.0.0"
 listen_addr_raft = "0.0.0.0"
 
 # The data dir hiqlite will store raft logs and state machine data in.
-# It must exist and be writable, unless you run a pure cache-only node (no
-# SQLite) with `cache_storage_disk = false`: that mode keeps everything
+# It must exist and be writable, unless hiqlite was built with the
+# `in-memory-snapshots` feature and runs a pure cache-only node (no SQLite)
+# with `cache_storage_disk = false`: that combination keeps everything
 # in-memory and never touches `data_dir`.
 # See also `cache_storage_disk`.
 #
@@ -145,8 +146,10 @@ wal_size = 2097152
 # all your cached data will be gone.
 # In-memory only hugegly increases the throughput though, so it
 # depends on your needs, what you should prefer.
-# In a pure cache-only deployment (no SQLite), `false` also means `data_dir`
-# is never used and does not need to exist.
+# With the `in-memory-snapshots` build feature, a pure cache-only deployment
+# (no SQLite) using `false` also keeps Snapshots in-memory, so `data_dir` is
+# never used and does not need to exist. Without that feature, Snapshots are
+# always written to `data_dir`.
 # See also `data_dir`.
 #
 # default: true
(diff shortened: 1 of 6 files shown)


sebadob:
This should not be the default, not even for in-memory nodes. Pushing the data into a File was a very intentional choice back then. When you keep it in memory, it doubles your memory cost. Memory is expensive, disk storage is cheap. And this snapshot is almost never needed.

Also, changing the type in the Raft declartion makes it impossible to stream the data from file in a very efficient way. It means that you would need to load the whole file into memory each time before you can even start sending it over the wire. While not a big deal for tiny caches, for big ones it's a pretty huge issue. 

If we actually want to make this possible, I would put it behind a feature flag that's opt-in and not even included in the `full` feature or anything like that. This choice should be made very intentionally, since it's really expensive.

When feature-gating it, we can also keep the old raft definition with the File, so we can stream efficiently by default, but switch to the other declaration only if it's set. I mean the current setup just uses the `data_dir` as a temp file storage when all the other parts are held in-memory, so almost no one would ever choose the "snapshot in-memory as well" option I guess.

I would probably call the feature something like `in-memory-snapshots`, and then very clearly document the cost that comes with it.

The only situation where this might be something people could use, is when their cache is tiny anyway, and adding a disk is just too expensive because they are running inside aws or something. But even then they could simply add no volume at all to the container and it would just use a temp file inside the container, which should work without issues even in that case.

bartekus:
That all makes sense, thanks for the detailed reasoning. You're right that doubling snapshot memory and losing zero-copy file streaming shouldn't be the default. I reworked it exactly along those lines:

- Added an opt-in `in-memory-snapshots` Cargo feature, requiring `cache` and deliberately **not** part of `full`.
- The default (feature off) is unchanged: `TypeConfigKV::SnapshotData` stays `tokio::fs::File`, snapshots are persisted to and streamed from `data_dir` as before, and `data_dir` is still required. The old Raft declaration and all the disk paths are kept as-is.
- Only with `in-memory-snapshots` does the declaration switch to `Cursor<Vec<u8>>`, and a memory-only node (`cache_storage_disk = false`) then keeps snapshots in memory and skips `data_dir` entirely. With the feature on but disk-backed, it still persists to disk and just loses the zero-copy streaming, which is the documented cost.
- Documented the cost on the feature itself (in `Cargo.toml`) and on `data_dir` / `cache_storage_disk` in both `NodeConfig` and `hiqlite.toml`.
- Refactored the shared serialization / persistence / apply logic into helpers so the two declarations don't duplicate it, added a `just clippy` line for the feature, and gated the unit test behind it.

So by default nothing changes; the in-memory option is purely additive for the niche case (tiny cache, no disk available) you described. Let me know if you'd rather name the feature differently or want it wired into the test matrix.

sebadob (approved):
Looks good now, thanks.

Took me a while to review. I had a few busy days, and with these changes it's easy to mess up something, so I wanted to not do it on the side.

bartekus:
It's all good sir, I'd add that precise things should never be rushed so you taking your time is not only expected; it is appreciated.

Thank you for all your ongoing hard work on these projects Sebadob, you rock!

PR #342, opened 2026-06-14
feat(config): resolve secret values from a separate secrets source

bartekus:
## What

Implements the `hiqlite` side of the external-secrets design from
sebadob/rauthy#1521: a secret-bearing config value can be set to the
case-sensitive sentinel `"$SECRETS"`, and the real value is then loaded by the
same key from a separate secrets source. This keeps the main config diffable and
version-controllable while secrets are managed separately (systemd
`LoadCredential`, Docker / Kubernetes secrets, ...).

This follows the shape you outlined in the issue.

## How

- `NodeConfig::from_toml` and `NodeConfig::from_toml_table` take a new
  `secrets: Option<toml::Table>` argument. If `Some`, it is the secrets source.
  If `None`, a `secrets_file` config option (or `HQL_SECRETS_FILE`) is honored as
  a fallback; it points to a TOML file that mirrors the config structure (the
  same `[hiqlite]` table).
- The sentinel is `"$SECRETS"`, case-sensitive. The allowed keys are exactly the
  ones you listed: `secret_raft`, `secret_api`, `s3_key`, `s3_secret`,
  `enc_keys`, `enc_key_active`, `password_dashboard`.
- Resolution happens inside the existing per-var parse helpers (`t_str` /
  `t_str_vec` get `_secret` variants), so the precise per-variable error messages
  are preserved, e.g. `secret_raft: "$SECRETS" set but "secret_raft" not found in
  secrets file`. Precedence per value is unchanged in spirit:
  env override > config value > `"$SECRETS"` resolution.
- `enc_keys` is the one array-typed secret: it is written as `enc_keys =
  "$SECRETS"` and resolved to the array stored under `enc_keys` in the secrets
  table.

## Breaking change

`from_toml` / `from_toml_table` gain the `secrets` argument (after
`table` / `table_name`, before the optional `enc_keys`). All in-tree callers
(examples, server, tests) are updated to pass `None`. Downstream callers
(including Rauthy's `from_toml_table` call) need the same one-line update.

## Tests

`hiqlite/src/config_toml.rs` gains unit tests for the resolution logic
(pass-through, sentinel resolved from table, missing key, no secrets source,
wrong type) plus integration tests for the `secrets_file` loading path
(missing file, structure mismatch). All green under the default `just test`
feature set.

## Docs

- `hiqlite.toml` documents `secrets_file` and the `"$SECRETS"` sentinel.
- `CHANGELOG.md` updated (incl. the breaking-change note).

### DIFF
--- a/hiqlite/src/config_toml.rs
+++ b/hiqlite/src/config_toml.rs
@@ -13,9 +13,23 @@ impl NodeConfig {
     ///
     /// You can overwrite most values from the file with ENV vars. If this is possible, it is
     /// mentioned in the documentation for each value.
+    ///
+    /// ## Secrets
+    ///
+    /// Secret-bearing values (`secret_raft`, `secret_api`, `s3_key`, `s3_secret`, `enc_keys`,
+    /// `enc_key_active`, `password_dashboard`) may be set to the case-sensitive sentinel
+    /// `"$SECRETS"`. In that case, the real value is looked up by the same key in a separate
+    /// secrets source. This keeps the main config diffable and version-controllable while the
+    /// secrets are managed separately (systemd `LoadCredential`, Docker / Kubernetes secrets, ...).
+    ///
+    /// The secrets source is, in order of precedence:
+    /// 1. the `secrets` table passed in here, if `Some`;
+    /// 2. otherwise a `secrets_file` config option (or the `HQL_SECRETS_FILE` env var) pointing to
+    ///    a TOML file that mirrors the config structure (i.e. holds the same `[{table}]` table).
     pub async fn from_toml(
         path: &str,
         table: Option<&str>,
+        secrets: Option<toml::Table>,
         #[cfg(any(feature = "s3", feature = "dashboard"))] enc_keys: Option<cryptr::EncKeys>,
     ) -> Result<Self, Error> {
         dotenvy::dotenv().ok();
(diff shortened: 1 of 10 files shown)


hiqlite/src/config_toml.rs
```diff
@@ -52,16 +67,48 @@ impl NodeConfig {
...
         dotenvy::dotenv().ok();
 
         let t_name = table_name;
         let mut map = table;
 
+        // Resolve the optional secrets source used by the `$SECRETS` sentinel. An explicitly
+        // passed-in `secrets` table wins; otherwise a `secrets_file` path (or `HQL_SECRETS_FILE`)
+        // is loaded, which must mirror the config structure (i.e. hold the same `[{t_name}]`
+        // table). Read up-front so the key is removed before the `check_empty` guard below.
+        let secrets_file = t_str(&mut map, t_name, "secrets_file", "HQL_SECRETS_FILE")?;
```
sebadob: It's unnecessary to read the value if `secrets == Some(_)` in the line below.

sebadob (changes requested):

hiqlite/src/config_toml.rs
bartekus: Good catch, addressed in c99365f: the `secrets_file` read now lives in the `None` arm, so it's only consulted when no `secrets` table is passed in.

sebadob (approved):
Thanks!

PR #351, opened 2026-08-11
fix(derive): Out-of-range from_i32 values error instead of being silently clamped

reneleonhardt:
## Summary

The `#[column(from_i32)]` attribute of the `FromRow` derive silently clamped `BIGINT` values to the `i32` range. A value like `5_000_000_000` was returned as `2_147_483_647` with no warning — a wrong result presented as a correct one.

This change makes the mapping fail loudly: out-of-range values now trigger a clear panic inside the generated `From` implementation, which surfaces to the caller as a query error through the existing `spawn_blocking` boundary instead of corrupting results.

## Changes

- `from_i32` (both plain and `Option` fields) now uses `i32::try_from(i64)` and panics with a descriptive message on overflow; the silent `cmp::min` / `cmp::max` clamping is removed.
- The derive internals return `proc_macro2::TokenStream` instead of `proc_macro::TokenStream` so the generated code can be unit-tested outside a macro-expansion context.
- Unit tests assert the generated code uses `TryFrom`, contains no clamping, and honors column renames.

## Verification

- `cargo test -p hiqlite-derive` — 2 passed.
- `cargo clippy -p hiqlite-derive -- -D warnings` — clean.
- `cargo test -p hiqlite --lib --features full` — 22 passed (generated code compiles in the full feature set).

## Compatibility notes

Behavior change: a column that previously silently clamped now errors. This is intentional. A fully fallible mapping (a `TryFromRow`-style trait) remains a future API decision; the existing data-triggered panics of `parse`, `flatten`, and plain column mapping are inherent to the infallible `From<&mut Row>` trait.

### DIFF
--- a/hiqlite-derive/src/from_row.rs
+++ b/hiqlite-derive/src/from_row.rs
@@ -2,7 +2,7 @@ use proc_macro2::{Literal, TokenStream, TokenTree};
 use quote::quote;
 use syn::{Attribute, Data, DeriveInput, Meta, MetaList, Type};
 
-pub fn impl_from_row(input: DeriveInput) -> proc_macro::TokenStream {
+pub fn impl_from_row(input: DeriveInput) -> proc_macro2::TokenStream {
     let name = input.ident;
     let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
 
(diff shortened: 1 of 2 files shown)


sebadob:
You pushed something that does not match the description at all.

reneleonhardt:
Sorry, human error, too many branches, this was the first low-effort one, but I failed even that 😅

sebadob:
Not sure if the pipeline will succeed. I have the suspicion that GitHub may have changed something to workers lately. I saw some timing issue with the tests pipeline lately when nothing in that regard was changed, so a failure in that case makes no sense. Let's see.

reneleonhardt:
You're not the only one, GitHub workers are extremely flaky at times, I see it in many repos.
If you can extract the most important checks into a nix build / devcontainer, I could run all of them locally to avoid burning your free quotas.
Actually I'm trying to run what I can see in Docker now, takes forever... and 7 more branches to go :(

sebadob:
If you have multiples ones, it makes sense. Workers are super slow as well.

If you `just clippy` and `just test-no-s3`, it should be fine.

sebadob:
You can of course run all of them in a container as well. The CI makes use of Rauthys builder container:

```
ghcr.io/sebadob/rauthy-builder:20260519
```

This has all the necessary tools installed.

sebadob:
The test failure now has nothing to do with your changes. I have never seen that as well though. Not sure how this can even happen. Probably only during a race condition. In such a case, the error could be handled gracefully, or we do another `get_raft_leader()`. If it still returns the same, this should be impossible to happen.

However, this PR looks fine, and it's independent from that one, thanks!

reneleonhardt:
AI found it, it's exactly one of 13 race fixes in the upcoming fix/dlock-handler-resilience 😅

sebadob:
Oh yeah the dlock has not been thoroughly tested yet. I added the feature because I thought I would need it, but never did so far. This is a very hard one. That's actually a very useful case for AI.

PR #352, opened 2026-08-11
fix: Make the distributed-lock handler fault-tolerant and deadlock-free

reneleonhardt:
## Summary

The distributed-lock subsystem (feature `dlock`) had several race conditions that could panic the lock handler — and with it every future lock on the node — or deadlock a lock key permanently:

- A lock holder slower than the 10 s lease caused an expiry takeover; the old holder's late `Drop` release then hit `panic!("... has been released already")`, killing the handler.
- A promoted waiter that died before re-acquiring left its ticket at the queue front forever; no new request could bypass it, so the lock key was blocked permanently.
- The client could orphan tickets by minting new ones on re-claim, and hit `unreachable!` when the handler granted a lock directly while it waited.

Related client/network hardening in the same area:

- Client requests now time out (120 s) instead of hanging forever on a stalled leader; a dead stream manager returns an error instead of panicking callers.
- The proxy relayed `LockAwait` failures into an `unimplemented!()` panic; the `Lock` response now carries the error and the client surfaces it.
- Corrupt frames from a cluster member drop the connection instead of panicking the raft client; dead panicking conversion impls were removed.

## Changes

- dlock handler: late/duplicate releases are ignored with a warning; awaits on removed locks are answered; acquire re-queues instead of asserting; stale queue-front tickets are drained once their lease window passes.
- dlock client: reuses its queue ticket across re-claims and accepts direct grants while waiting; lock waits are bounded by the request timeout.
- Client stream manager: 14 response awaits share a 120 s timeout helper; closed channels and stalls return errors.
- Proxy: `LockAwait` errors propagate through the response payload.
- Raft network: corrupt member frames drop the connection; dead `From` impls removed; leader-membership lookup errors instead of panicking.

## Verification

- `cargo test -p hiqlite --lib --features full` — 31 passed, including 6 lock handler tests (protocol roundtrip, duplicate/late release, removal races) and 3 timeout-helper tests.
- `TEST_SKIP_S3_RESTORE=true cargo test --features cache,counters,dlock,listen_notify,macros,toml` (CI matrix, lib+bins) — 31 passed; `cargo test -p hiqlite-wal` — 7 passed.

## Compatibility notes

Lock lease semantics are unchanged (leader wall clock, 10 s lease); the deterministic-timestamp alternative requires a breaking raft entry format change and is intentionally left as a documented design decision. The proxy `Lock` response payload changes shape (now `Result`) — client and server ship in the same version.

### DIFF
--- a/hiqlite/src/client/backup.rs
+++ b/hiqlite/src/client/backup.rs
@@ -1,3 +1,4 @@
+use crate::client::helpers::await_channel_response;
 use crate::client::stream::{ClientBackupPayload, ClientStreamReq};
 use crate::network::api::ApiStreamResponsePayload;
 use crate::store::state_machine::sqlite::state_machine::QueryWrite;
@@ -94,9 +95,7 @@ impl Client {
                 }))
                 .await
                 .map_err(|err| Error::Error(err.to_string().into()))?;
-            let res = rx
-                .await
-                .map_err(|_| Error::Error("Client stream manager closed".into()))??;
+            let res = await_channel_response(rx).await??;
             match res {
                 ApiStreamResponsePayload::Backup(res) => res,
                 _ => unreachable!(),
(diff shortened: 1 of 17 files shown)


sebadob (approved):
Thanks!

PR #353, opened 2026-08-11
fix(hiqlite-wal): Oversized raft entries panic by default, opt into error

reneleonhardt:
## Summary

A single raft log entry cannot span WAL files. Entries larger than `wal_size` (the default is 2 MB — easily reached by one large `INSERT`, transaction, or batch) previously `panic!`ed inside the WAL writer thread, silently killing the log store and taking the node down. Per maintainer review, an oversized entry is a non-recoverable setup issue (config change with full restart, or code changes), so the writer panics by default again; the new `oversized-entry-error` feature switches the append to fail with the dedicated `Error::WalSizeExceeded` variant instead, so it can be caught properly and the writer keeps serving subsequent requests.

## Changes

- `hiqlite-wal` feature `oversized-entry-error` (off by default): when enabled, an oversized append fails with the new `Error::WalSizeExceeded` variant and the writer continues; the panic is the default behavior.
- Regression tests cover both paths: with the feature, a 64 KB WAL + 70 KB entry errors and the writer survives a normal append; without it, the oversized entry panics the writer thread and no further append is acked.

## Verification

- `cargo test -p hiqlite-wal` — 8 passed (panic default, incl. the writer-dies test).
- `cargo test -p hiqlite-wal --features oversized-entry-error` — 8 passed (incl. the error-and-survive test).
- `cargo clippy -p hiqlite-wal --all-features -- -D warnings` — clean.
- `cargo check -p hiqlite --features full` — clean (consumer builds against the new variant).

## Compatibility notes

Default behavior is unchanged from upstream: an oversized entry still panics the writer, deterministically on every member. Enabling `hiqlite-wal/oversized-entry-error` makes the append fail with `Error::WalSizeExceeded` instead, which callers can catch; the write is then rejected deterministically on every member. Dynamically growing WAL files to accept arbitrarily large entries remains a possible future feature and is explicitly not attempted here.

### DIFF
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -191,11 +191,35 @@ fn run(
                     let mut active = wal.active();
                     while let Ok(Some((id, bytes))) = rx.recv() {
                         if bytes.len() > data_len_limit {
-                            panic!(
-                                "`data` length must not exceed `wal_size` -> data length is {} \
-                            vs wal_size (without header) is {data_len_limit}",
-                                bytes.len(),
-                            );
+                            // A single raft entry cannot span WAL files. By default an
+                            // oversized entry is a non-recoverable setup issue (it needs a
+                            // config change with a full restart, or code changes), so the
+                            // writer panics. With the `oversized-entry-error` feature the
+                            // append instead fails with `Error::WalSizeExceeded` and the
+                            // writer keeps serving subsequent requests. With the default
+                            // `wal_size` of 2MB this is easily reached by a single large
+                            // INSERT, transaction or batch.
+                            #[cfg(feature = "oversized-entry-error")]
+                            {
+                                res = Err(Error::WalSizeExceeded(
+                                    format!(
+                                        "`data` length must not exceed `wal_size` -> data \
+                                        length is {} vs wal_size (without header) is \
+                                        {data_len_limit}",
+                                        bytes.len(),
+                                    )
+                                    .into(),
+                                ));
+                                break;
+                            }
+                            #[cfg(not(feature = "oversized-entry-error"))]
+                            {
+                                panic!(
+                                    "`data` length must not exceed `wal_size` -> data length \
+                                    is {} vs wal_size (without header) is {data_len_limit}",
+                                    bytes.len(),
+                                );
+                            }
                         }
 
                         if !active.has_space(bytes.len() as u32) {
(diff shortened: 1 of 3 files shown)


hiqlite-wal/src/writer.rs
```diff
@@ -191,11 +191,20 @@ fn run(
...
-                                "`data` length must not exceed `wal_size` -> data length is {} \
-                            vs wal_size (without header) is {data_len_limit}",
-                                bytes.len(),
-                            );
+                            // A single raft entry cannot span WAL files. Fail the append with a
+                            // regular error instead of panicking: a panic kills the whole WAL
+                            // writer thread and with it the log store, taking the node down.
+                            // With the default `wal_size` of 2MB this is easily reached by a
+                            // single large INSERT, transaction or batch.
+                            res = Err(Error::Generic(
```
sebadob: I actually prefer the `panic`, because if this happens, you have a general issue with your whole setup, and this is not recoverable. It needs either config changes with a full restart, or actual code changes. 

However, we can return an error here, but I would at least create a new variant like `Error::WalSizeExceeded` or something like that, so it can be caught properly. Or, even better, an additional feature that switches from `panic` to error in such a case. I don't really care if `panic` is the default, or if we need to set a feature flag to make it panic.

sebadob (changes requested):

hiqlite-wal/src/writer.rs
reneleonhardt: Complied. And you really want to double all your builds into -error and -panic variants instead of supporting a new config?

sebadob (approved):

hiqlite-wal/src/writer.rs
sebadob: > And you really want to double all your builds into -error and -panic variants instead of supporting a new config?

No, I don't. I keep it very simple: `panic` for non-recoverable errors and handle everything gracefully in all other cases. So, I would never use the error version in this case. If this ever happens, either the app logic has serious issues, or the config is very far off. Both things should never happen.

sebadob: Btw I thought about testing a non-memmapped version of this, if there is any real difference in performance. If I can make it work without memmapping, at least the writer, it would allow for dynamic file growth, which would make sure that this edge case can never happen. I assume that this will cost quite a bit of performance though, not sure.

PR #354, opened 2026-08-11
fix(dashboard): Throttle login attempts per client

reneleonhardt:
## Summary

The dashboard login proof-of-work is currently disabled: loading the spow WASM in the embedded Svelte 5 build failed during the framework's preview era (sveltejs/svelte#13155, closed without a fix; TODO in `Login.svelte`). Re-enabling it is tracked separately as `feat/dashboard-pow-reenable` — the import now builds, hydrates and computes on the stable stack (Svelte 5.56.8 / Vite 8.2 / SvelteKit 2.70.2), but the server-side `Pow::validate` and the secure-context (TLS) requirement noted in `handlers.rs` are addressed there, not here. With PoW gone, the login endpoint was freely brute-forceable; the only protection was a global single-flight lock on password verification, which caps parallelism but not the total attempt rate.

This change adds per-client failed-login throttling as the active protection.

## Changes

- Failed logins are tracked per client, keyed by the rightmost `X-Forwarded-For` hop (the entry added by the closest proxy).
- After 5 failures the client is locked out with `429 Too Many Requests` for 60 seconds; the check runs before password verification so locked-out clients do not consume argon2 work.
- A lockout expiry resets the counter, so a legitimate user is never locked forever by a single extra failure.
- Clients without a proxy header share no bucket — a spoofed or missing header cannot be used to lock everyone out.
- Successful login clears the failure counter.
- The existing global single-flight lock in password verification remains as the throughput backstop.

## Verification

- `cargo test -p hiqlite --lib --features full` — 36 passed, including 5 new login-throttle unit tests (lock after max failures, expiry reset, clear on success, bounded map, rightmost-header keying).

## Compatibility notes

The PoW challenge endpoint and the frontend PoW plumbing stay in place; the throttle is the active protection. The WASM-loading blocker from the Svelte 5 preview era is gone — verified on Svelte 5.56.8 / Vite 8.2 / SvelteKit 2.70.2: the spow import builds, hydrates, and `pow_work_wasm` computes a proof in the browser. Re-enabling PoW (`feat/dashboard-pow-reenable`) re-activates `Pow::validate` in `post_session` (the spow secret is already initialized from the configured encryption key at startup) and serves the dashboard over TLS via `tls_api` / `tls_auto_certificates` for the secure context.

### DIFF
--- a/hiqlite/src/dashboard/session.rs
+++ b/hiqlite/src/dashboard/session.rs
@@ -12,8 +12,10 @@ use chrono::Utc;
 use cryptr::EncValue;
 use cryptr::utils::{b64_decode, b64_encode};
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::env;
-use std::sync::LazyLock;
+use std::sync::{LazyLock, Mutex, OnceLock};
+use std::time::{Duration, Instant};
 use tracing::debug;
 
 const COOKIE_NAME: &str = "__Host-Hiqlite-Session";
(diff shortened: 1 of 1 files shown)


sebadob (changes requested):
That is not really a solution. This is purely based on the existence of an `x-forwarded-for` header. If it does not exist, there is no rate limit (instead of using the direct peer IP). This means it will only work when running behind a reverse proxy. 

Another issue is that this header can be easily spoofed, because you cannot know if you can trust the header.

There is also no cleanup. This means an attacker could abuse it for resource exhaustion, making lots of requests with endless amounts of spoofed IPs, and fill up the memory, even though it would take quite a while.

Data is also not synced between nodes. When you run behind a reverse proxy, you will most probably load balance these requests, but these lockouts are not synced.

Doing this properly is a lot more work, and needs more configuration upfront, so you know if you can trust your direct peer IP (proxy).

Because of all of these reasons, and because the dashboard is not really meant for public exposure, I added the PoW. It's stateless, it does not care about all of this additional setup, checks, and syncing. The reason why it is still disabled is because it does not work over plain HTTP, which is the reason why the WASM did not load. But, the TODO and the disabling was from a time when hiqlite did not have the auto-TLS feature. Enabling TLS is a no brainer now.

---

Long story short, I would make this a lot simpler. I would restrict logins to "only 1 at a time", and then write a 5 second timeout into some global value to mark the next timestamp when another login can be done, completely independent from any IP. This slows everything down so much that it negates any brute force attempts. There is no need for concurrent logins to this dashboard.

This would require a tiny update to the UI part, so it will lock the input until the next try of course. Another solution would be to still accept logins during lockout, but do a short `sleep` until `exp`, and only then proceed hashing the password. When we only ever add 5-10 seconds, the UI will not timeout in such a short time anyway, it will just take longer after a failure.

Additionally, I would dynamically require a PoW again. If we are serving via HTTPS, require a PoW, and ignore it via HTTP. This can be checked independently on both sides, backend and UI very easily.

This solution does not sync the lockouts as well, but we do have the PoW in almost all situations in addition. Even when you would run a 3 node HA cluster and you could to 3 times the requests (because of no lock sync), it would be impossible to brute force any meaningful password.

reneleonhardt:
I'm closing this branch, and moving the global 5s cooldown + UI lock to the already prepared feat/dashboard-pow-reenable.

PR #355, opened 2026-08-12
feat(dashboard): Re-enable login PoW over TLS and add a global login cooldown

reneleonhardt:
## Summary

The dashboard login endpoint had no effective anti-brute-force protection: the proof-of-work was disabled (a Svelte 5 preview-era WASM loading failure, sveltejs/svelte#13155, closed without a fix), and the per-client throttle from the previous PR was withdrawn after review (spoofable `X-Forwarded-For`, no cleanup, no cross-node sync). This PR restores PoW as the primary, stateless protection — required only when the dashboard is served over TLS — and adds a global, IP-independent 5-second cooldown after failed logins as defense in depth.

## Changes

- Global login cooldown: a failed login sets a 5-second cooldown before the next attempt, tracked in a single global value with no per-client state to spoof or exhaust; the UI locks the form until it expires. The existing single-flight lock on password hashing is unchanged.
- Proof-of-work re-enabled: the spow WASM flow in the login form and the server-side `Pow::validate` in `post_session`, verified against the per-node secret initialized at startup from the active encryption key.
- PoW required only over TLS: validated when the API server listens on TLS (auto-TLS via `tls_api` / `tls_auto_certificates`), ignored over plain HTTP; the browser mirrors this via `window.isSecureContext`.

## Toolchain

The dashboard deps are updated to their newest releases (npm-check-updates: @sveltejs/vite-plugin-svelte 7.2, @swc/core 1.15, rollup 4.62, svelte-check 4.7). The four runtime-critical packages stay pinned to the verified versions (svelte 5.56.8 / vite 8.2.0 / SvelteKit 2.70.2 / adapter-static 3.0.10). TypeScript 7 (native compiler) is opted into via svelte-check's documented `--tsgo` setup (the `@typescript/native` alias plus the `--tsgo` flag, with typescript@6 as the fallback) and is isolated in its own commit so it can be dropped without affecting the other updates.

The wasm-bindgen glue in `src/spow/` is generated code; with `checkJs` enabled it produced ~76 implicit-any findings. It is excluded from the type-checked program, which is the only correct fix: the file cannot be hand-fixed (wasm-pack regenerates it), there is no per-directory `checkJs` option, and `@ts-nocheck` would be silently lost on regeneration. The exclude mirrors the svelte-kit tsconfig's own excludes (TypeScript does not merge `exclude` across `extends`) plus `src/spow/**`; the ambient `spow-wasm.d.ts` stays in the program, so the `pow_work_wasm` import keeps its types.

## Verification

- `cargo test -p hiqlite --lib --features full` — 33 passed (includes the cooldown test and the PoW challenge/work/validate roundtrip).
- Dashboard deps pinned to the verified toolchain (svelte 5.56.8 / vite 8.2.0 / SvelteKit 2.70.2 / adapter-static 3.0.10).
- `npm run check` (svelte-check) — 0 errors, 0 warnings (the generated spow glue is excluded from the checked program; see Toolchain). `npm run test:smoke` (Playwright, headless chromium) — 1 passed: the login page builds, loads, and hydrates without JS errors. `pow_work_wasm` was additionally exercised in a headless browser and computes a proof.

## Compatibility notes

Over TLS (auto-TLS enabled), logins require a valid PoW proof; over plain HTTP the proof is skipped and the global cooldown is the protection. Behind a TLS-terminating reverse proxy the browser sends a proof while a plain-HTTP backend ignores it — no rejection either way.


dashboard/src/lib/components/Login.svelte
```diff
@@ -4,41 +4,42 @@
     import InputPassword from "$lib/components/form/InputPassword.svelte";
     import {API_PREFIX, fetchGet} from "$lib/utils/fetch";
     import {storeSession} from "$lib/stores/session";
-    // import {pow_work_wasm} from "../../spow/spow-wasm";
+    import {pow_work_wasm} from "../../spow/spow-wasm";
```
sebadob: Have you tested this over plain HTTP? I remember this failing, because WASM requires TLS. When this is a global import, it crashes the JS runtime afaik, so we need a dynamic import only when served via HTTPS.

If this produces issues, we could also copy & paste the "new and better" way of doing PoW in the browser from Rauthy. It starts a worker dynamically only when necessary. This also does not block the JS event loop. The PoW impl for this dashboard is very simple.

The worker def is here: https://github.com/sebadob/rauthy/blob/main/frontend/src/workers/pow.ts

And then there is only this function to do everything to solve it: https://github.com/sebadob/rauthy/blob/main/frontend/src/utils/pow.ts

You should be able to do a simple dynamic import below in the `if (window.isSecureContext)` as well. As long as any animations like loading spinner and so on are set before solving the PoW, nothing will be blocked. Only inputs and so on, but the user does not need to do anything anyway while waiting for the pow. This would be a simpler solution.

dashboard/src/lib/components/Login.svelte
```diff
@@ -4,41 +4,42 @@
...
     const action = `${API_PREFIX}/session`;
 
     let error = $state('');
     let isLoading = $state(false);
+    let lockUntil = $state(0);
+    let locked = $derived(Date.now() < lockUntil);
 
-    async function onSubmit(form: FormData, params: URLSearchParams) {
+    async function onSubmit(form: HTMLFormElement, params: URLSearchParams) {
+        if (locked) return;
```
sebadob: Even though this should never happen as long as the button disabled properly, we should maybe at least log an error to the console, or even show one to the user. When the input is not locked, which makes sense, so the user can update before the unlock, it should be possible to trigger the submit when pressing enter. So, it would be possible to get here.

dashboard/src/lib/components/Login.svelte
```diff
@@ -53,6 +54,13 @@
             storeSession.set(resp);
         } else {
             error = Object.values(resp)[0] as string;
+            // Lock the form for the global login cooldown after a 429.
+            if (res.status === 429) {
+                lockUntil = Date.now() + 5000;
```
sebadob: This is a bit brittle. If the constant in the backend is changed, this is lost very easily. It would be better to add a `Retry-After: <delay-seconds>` header and take the value from that one.

dashboard/src/lib/components/Options.svelte
```diff
@@ -49,10 +49,19 @@
     });
 
     let refOptions: undefined | HTMLElement = $state();
-    let usePopover = $state(fallbackOptions ? false : asPopover);
+    let usePopover = $state<boolean | undefined>(false);
     let close: undefined | (() => void) = $state();
 
-    let selected = $state(withSearch ? -1 : 0);
```
sebadob, line 55: Why changing the logic for a few of these initial values? Is it because of these (stupid) svelte linting error that it catches only the initial value?

I would love to have a way to disable these completely, but you can get rid of them by wrapping the initalization in `untrack(() => {})`. This lint hurts readability, and is just super annoying. It's very clear that this of course only catches the initial value, but at least last time I checked on it, they were against adding an option to disable this.

dashboard/src/lib/components/Tabs.svelte
```diff
@@ -15,9 +15,11 @@
         width?: string,
     } = $props();
 
-    if (tabs.length > 0 && selected === '') {
-        selected = tabs[0];
-    }
```
sebadob, line 20: Same with the stupid inital value linting here I guess? You can use `untrack` again. Not that big of a deal here, but this only ever needs to run once.

dashboard/package.json
```diff
@@ -6,20 +6,23 @@
...
+    "@sveltejs/adapter-static": "3.0.10",
+    "@sveltejs/kit": "2.70.2",
+    "@sveltejs/vite-plugin-svelte": "^7.2.0",
+    "@swc/core": "^1.15.47",
+    "@typescript/native": "npm:typescript@^7.0.2",
     "esbuild": "^0.28.1",
-    "rollup": "^4.60.4",
-    "svelte": "^5.55.7",
-    "svelte-check": "^4.0.5",
+    "rollup": "^4.62.4",
```
sebadob: When the WASM import is done dynamically, either via direct import or WebWorker, it should be possible to drop `rollup` and also `top-level-await` here.

hiqlite/src/dashboard/mod.rs
```diff
@@ -4,8 +4,23 @@ use cryptr::EncKeys;
...
+/// from `NodeConfig::tls_api`; the login proof-of-work is required only in that case,
+/// because the WASM client needs a secure context. The browser mirrors this via
+/// `window.isSecureContext`.
+static API_TLS_ENABLED: AtomicBool = AtomicBool::new(false);
+
+pub fn set_api_tls(enabled: bool) {
+    API_TLS_ENABLED.store(enabled, Ordering::Relaxed);
+}
+
+pub fn api_tls_enabled() -> bool {
```
sebadob: Tiny thing, but I like to have all `bool`s prefixed with an `is` or `has`.

hiqlite/src/dashboard/session.rs
```diff
@@ -27,6 +29,29 @@ pub static INSECURE_COOKIES: LazyLock<bool> = LazyLock::new(|| {
         .expect("Cannot parse HQL_INSECURE_COOKIE as bool")
 });
 
+/// Global login cooldown after a failed attempt, independent of any client identity:
+/// the dashboard is an ops surface with no need for concurrent logins, so after a
+/// failure the next login is rejected for a few seconds. There is no per-client state
+/// to spoof or exhaust, and the single-flight lock in `password::verify_password`
+/// still serializes the actual hashing.
+const LOGIN_COOLDOWN: Duration = Duration::from_secs(5);
```
sebadob, line 38: This constant should be forwarded to the UI with either a `429` or after `401`, so we only ever need to update in a single location if necessary.

hiqlite/src/dashboard/session.rs
```diff
@@ -27,6 +29,29 @@ pub static INSECURE_COOKIES: LazyLock<bool> = LazyLock::new(|| {
...
+static NEXT_LOGIN_ALLOWED: Mutex<Option<Instant>> = Mutex::new(None);
+
+fn login_locked() -> bool {
```
sebadob: Same as above for `bool`s, please prefix with `is_`.

sebadob (changes requested):

reneleonhardt:
All comments addressed:

- **C1** (Login.svelte:7 — dynamic import): done — WASM imported only inside `window.isSecureContext`; verified over plain HTTP (non-localhost origin, `isSecureContext` false, 0 JS errors) and in a secure context (proof computes).
- **C2** (locked submit): a locked submit now shows the remaining wait instead of silently returning.
- **C3/C8** (Retry-After): the 429 carries `Retry-After` from `LOGIN_COOLDOWN` (remaining wait); the UI reads it, no hardcoded 5 s.
- **C4/C5** (untrack): Options/Tabs initial-value logic restored and wrapped in `untrack`.
- **C6** (drop rollup + top-level-await): removed; build/check/smoke green without them.
- **C7/C9** (is_ prefixes): boolean identifiers renamed (`is_api_tls_enabled`, `is_login_locked`, `is_enabled`).

Beyond the comments, also changed (from codemap review and testing):
- `LoginRequest.pow` is `#[serde(default)]` — plain-HTTP submissions carry no `pow` and previously failed with a 400.
- PoW-TLS gating and the cooldown 429 extracted into unit-tested seams.
- Pre-existing cleanup kept: svelte-check 0/0 (via `src/spow/` exclusion + component type fixes), Playwright smoke test, TS7 opt-in as a droppable commit.

sebadob (approved):
Thanks!

PR #356, opened 2026-08-12
fix(cache): Make TTL expiry collision-safe and refresh-aware

reneleonhardt:
## Summary

- Keys expiring in the same second overwrote each other in the expiry map — one would never expire and stay stale forever.
- Re-putting a key with a new TTL kept the old expiry registered — the fresh value was deleted early.
- Re-putting a value without a TTL left a stale expiry that could delete the now-eternal value.

## Changes

- TTL handler keeps a `key -> expiry` index and an `expiry -> keys` map: multiple keys may share an expiry second; refreshing a key removes its old expiry in O(1); a `Clear` request removes a pending expiry when a value is re-put without a TTL.
- The expiry-delete guard compares against the absolute bucket time, so a refreshed key is never deleted at its old expiry.
- A pending Ttl/Clear request is processed before an expiring bucket, so a refresh landing in the same instant cannot let a stale Delete hit the freshly re-put value.
- The cluster test's read-after-insert assertions now wait for replication, so they cannot flake under CI load.

## Verification

- `cargo test -p hiqlite --lib --features full` — 36 passed on the rebased base (cdce0f7), including 5 TTL handler tests, all deterministic via an injected clock (same-second expiry, refresh not deleted early, refresh expires at the new time, clear, snapshot roundtrip).

## Compatibility notes

Snapshot format is unchanged. The snapshot stores one key per expiry timestamp, so if two keys expire in the same second, a snapshot restore keeps only one of their TTLs — a rare edge case, documented in the handler, preferable to dropping TTLs during live operation.

### DIFF
--- a/hiqlite/src/client/cache.rs
+++ b/hiqlite/src/client/cache.rs
@@ -216,7 +216,7 @@ impl Client {
                 cache_idx: cache.hiqlite_cache_index(),
                 key: key.into(),
                 value,
-                expires: ttl.map(|seconds| Utc::now().timestamp().saturating_add(seconds)),
+                expires: ttl.map(|seconds| Utc::now().timestamp_micros().saturating_add(seconds.saturating_mul(1_000_000))),
             },
             false,
         )
--- a/hiqlite/src/store/state_machine/memory/cache_ttl_handler.rs
+++ b/hiqlite/src/store/state_machine/memory/cache_ttl_handler.rs
@@ -1,7 +1,7 @@
 use crate::store::state_machine::memory::kv_handler::CacheRequestHandler;
 use crate::store::state_machine::memory::state_machine::{CacheResponse, StateMachineData};
 use chrono::Utc;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{RwLock, oneshot};
@@ -11,6 +11,9 @@ use tracing::{debug, warn};
 #[derive(Debug)]
 pub enum TtlRequest {
     Ttl((i64, String)),
+    /// Removes a previously registered expiry for a key, e.g. when the value was re-put without
+    /// a TTL. Without this, the old expiry would still fire and delete the fresh value early.
+    Clear(String),
     SnapshotBuild(oneshot::Sender<BTreeMap<i64, String>>),
     SnapshotInstall((BTreeMap<i64, String>, oneshot::Sender<()>)),
 }
@@ -22,7 +25,13 @@ pub fn spawn(tx_kv: flume::Sender<CacheRequestHandler>) -> flume::Sender<TtlRequ
 }
 
 async fn ttl_handler(tx_kv: flume::Sender<CacheRequestHandler>, rx: flume::Receiver<TtlRequest>) {
-    let mut data: BTreeMap<i64, String> = BTreeMap::new();
+    // expiry timestamp -> keys expiring at that time. A `Vec` is used because multiple keys can
+    // expire in the same second. A map keyed by expiry alone would silently drop all but one of
+    // them, so the remaining ones would never be expired.
+    let mut data: BTreeMap<i64, Vec<String>> = BTreeMap::new();
+    // key -> currently registered expiry. Needed to remove an old expiry in O(1) when a value is
+    // refreshed, so a stale expiry can never delete a freshly updated value early.
+    let mut exp_of: HashMap<String, i64> = HashMap::new();
 
     loop {
         let sleep_exp = {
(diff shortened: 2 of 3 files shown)


sebadob (changes requested):
This is a rather wasteful implementation. 

We have a additional memory allocations for each entry to `data`, and we have the additional map with a clone of each key, and another clone of the exp. This requires basically double the resources, and is more complicated than it needs to be.  That's why the TODO was still in the code. 

I took a look at the current code, and I am not sure why the TODO mentions microseconds precision currently. I think this is a leftover. I migrated to `chrono` quite a while ago, and I think the other TS mechanism I used provided a micros TS by default, not sure anymore.

However, to get rid of possible collisions, there is a simpler and more efficient solution:

- Update `hiqlite/src/client/cache.rs` -> `put_bytes()` and instead of using second precision when calculating `expires`, change it to micros. This already reduces the chance of a collision by 1.000.000.
- During `TtlRequest::Ttl`, simply check if there is an entry with the exact same value, and if so, add a single microsecond to the new one until there is a free space. It's very likely that the next one is free already, and even if not, iterating a few times is cheap.
- Get rid of the `Vec<_>` as the value for our cache. Collisions cannot happen anymore and this removes another syscall for each put.

The precision could be changed to nanos as well to have another factor of 1000 of safety margin. However, this would double the cost for the timestamp in all locations from an `i64` to `i128`. 4 byte is not much on its own, but this will add up, especially for caches that are huge and might produce a collision.

Since the full TS is caluclated on the client side before the message is pushed into the Raft, and as long as each cluster node uses the same algorithm to resolve collisions, consistency is guaranteed.

This leaves is with one tiny "issue". This is a breaking change, and it would make it incompatible with currently running caches when starting the new version. To resolve this, I would not go with nanos (uses more memory anyway). This keeps the `exp` an `i64`. We can guarantee compat with older cache entries by checking the exp size when we calculate the `first_exp` and multiply any possibly older seconds value. This check could be removed a couple of versions later.

When we do it this way, we don't need the additional HashMap, no `TtlRequest::Clear`, no `exp_of` checks and inserts. The only situation when this algorithm will break is when you consistently insert 1.000.000 keys with the exact same TTL each second without any gaps into the same cache (with micros precision). When you are running at this scale, you will not run an embedded cache anymore, but a full separate server for caching only. So I would say we are safe.

---

Apart from this, when taking a look at this old code, I would do it differently today. Most often, a BTreeMap is faster for integer keys, which is why it's not a HashMap currently. However, I now would either use a HashMap with `ahash` hasher, or even better, the default one and a custom identity hasher for `i64`, since we don't even need to do any hashing in this case. The identity hasher is preferred to avoid an additional dependency, but this can easily be another PR. This would need to be done for all the other in-memory caches as well. 

A HashMap wouild also allow us to pre-allocate some capacity, and even provide the user to provide a custom one for each cache via the `CacheVariants` trait.

reneleonhardt:
Understood.
One gap in the micros design: re-putting a key with a new TTL leaves the old expiry entry in the map, so it fires later and deletes the fresh value early. Micros fixes collisions, not stale entries.

Options:
- accept the behavior change (re-put no longer extends the TTL),
- or keep a minimal per-key index to drop the old expiry on re-put.
 
Which do you prefer?

sebadob:
> One gap in the micros design: re-putting a key with a new TTL leaves the old expiry entry in the map

Oh, yes makes sense of course. That makes it a bit more tricky again. The expected behavior would be that a re-PUT extends the TTL, or basically resets it. In that case, we would actually need a map for reverse lookup I guess, sadly. 

In that case though, I am thinking about adding `ahash` and use their hasher. It's a lot faster than the default one, but it's not brute force resistant anymore like the default one. This is fine in lots of situations, but can become an issue if user-controlled input is used as cache keys.

Maybe we should keep the default `HashMap` for now, just to be secure by default, and then add a feature flag to opt-in to `ahash` for performance. But this would be another PR.

When we need to keep the additional reverse lookup index though, we could also keep the collision resistance checks like they are now again. We could also have a channel from the KV handler to the TTL handler and inform it about an existing key, but without the reversed map, we would need to scan all existing entries with `O(n)`, which would be super bad. However, we could also update the `kv_handler` in a way that it does not only save the `Vec<u8>` data, but also the `Option<i64>` exp timestamp instead. Then it could inform the TTL handler again. This would save use an additional `String` overhead for each entry, and the whole reverse lookup map, but at the cost at a lot more complexity.
Sorry, this was just me thinking loudly.

So I guess keeping the additional map in the TTL handler is actually the best soltution, you're right.

In terms of collision resistance, we could keep a `Vec<_>` there and have it guaranteed robust, but I think micros precision is way good enough and it saves us some additional syscalls.

reneleonhardt:
Implemented your suggestion:

- Expiries now **microseconds** with a **1µs bump** on collision — the `Vec` is gone.
- Re-putting a key **extends its TTL** via the kept `key → expiry` index; `Clear` stays.
- Old second-precision expiries from previous snapshots are **normalized on load**.
- Tests now drive the handler with an **injected clock** (deterministic, no sleeps).
- One commit on top of the previously reviewed history; the HashMap/ahash switch stays a separate PR.

reneleonhardt:
- HashMap follow-up will be: identity hasher for the i64 expiry map + default HashMap for the String-keyed cache maps (secure by default, as you suggested).
- The ahash opt-in is tiny: a default-off feature flag swapping only the hasher on the String-keyed maps (~30–50 lines, no overlap, stacks on top).
- Does a commit that small need its own third branch/PR, or can it ride on the HashMap PR / come as a single follow-up commit?

sebadob:
> Does a commit that small need its own third branch/PR, or can it ride on the HashMap PR / come as a single follow-up commit?

Usually not, but it would be cleaner, since there are probably a few other location where using it makes sense. Also, maybe switching some others apart from the KV handler to identity hashers makes sense. I am pretty sure at least the maps for inflight requests in cluster communication is using an integer as key as well, and in some other locations it's probably a big win if we change to `ahash`. So this other PR could do a global check for any `HashMap::new()` or `with_capacity()` and see, if we can switch the hasher.

Another idea could be that we add a `fn capacity(&self)` for the `CacheVariants` trait. By default, it could return a small number like `8` for all entries, and for optimization, the user could overwrite it to return better values that fit the application for upfront mem alloc. It makes sense to combine these changes in a PR. Then it's basically only about perf opt.

sebadob (approved):
Thanks

PR #357, opened 2026-08-12
fix(hiqlite-wal, cache, sqlite): Write WAL metadata, snapshots, and backups crash-atomically

reneleonhardt:
## Summary

Several durable files were written non-atomically, so a crash in the middle could leave a torn or missing file that recovery then misinterprets:

- The WAL metadata file was removed and recreated in place with no fsync. A crash between the two operations left `meta.hql` missing; on restart the node forgot its persisted raft vote (a safety-relevant window) and its purged-log frontier.
- Cache snapshots were copied to their final path, so a crash mid-copy left a partial snapshot that could be read as valid.
- SQLite snapshots were vacuumed directly to their final name, and incoming snapshots were copied into place.
- Backups were vacuumed directly under their final name, so a crash could leave a partial file a later restore would pick up as healthy.

All of these now use the write-temp-file, fsync, atomic-rename pattern. Additionally, S3 backup uploads are retried instead of logged once, because the backup request already acked before the async upload ran.

## Changes

- WAL `meta.hql`: temp file + `sync_all` + `rename`, with a best-effort directory fsync; non-unix keeps the previous remove-then-rename behavior (std `rename` refuses to overwrite there).
- Cache snapshot persist: temp file + fsync + rename; stale temp files no longer block the next snapshot; the old-snapshot cleanup task warns on read/remove errors instead of panicking.
- SQLite snapshot create and install: `VACUUM INTO` a temp file then rename; install renames instead of copying.
- Backups: `VACUUM INTO` a temp file then rename; failed S3 pushes retry up to 3 times with backoff and fail loudly.

## Verification

- `cargo test -p hiqlite-wal` — 8 passed (includes a new atomic-overwrite test covering the smaller-payload case that previously broke in-place overwrites).
- `cargo test -p hiqlite --lib --features full` — 22 passed (branch base; re-verify on rebase).

## Compatibility notes

The metadata file format is unchanged; only the write procedure differs. A persistent S3 outage still loses the offsite copy once the local backup passes the retention window (documented in the commit).

### DIFF
--- a/hiqlite-wal/src/metadata.rs
+++ b/hiqlite-wal/src/metadata.rs
@@ -70,7 +70,7 @@ impl Metadata {
     #[inline]
     fn write_unchecked(bytes: &[u8], base_path: &str) -> Result<(), Error> {
         let path = format!("{base_path}/meta.hql");
-        let tmp_path = format!("{path}.tmp");
+        let tmp_path = format!("{path}~");
 
         let mut file = File::create(&tmp_path)?;
 
@@ -84,9 +84,8 @@ impl Metadata {
         file.sync_all()?;
         drop(file);
 
-        // atomic replace: readers/crash recovery never see a torn or missing meta file
-        #[cfg(not(unix))]
-        let _ = fs::remove_file(&path);
+        // atomic replace: the rename alone publishes the new file, so readers/crash
+        // recovery never see a torn or missing meta file
         fs::rename(&tmp_path, &path)?;
 
         // best-effort: make the rename itself durable across power loss
@@ -126,8 +125,8 @@ mod tests {
     }
 
     #[test]
-    fn metadata_overwrite_is_atomic() -> Result<(), Error> {
-        let base_path = format!("{}/metadata_overwrite_is_atomic", PATH);
+    fn metadata_overwrite_replaces_existing() -> Result<(), Error> {
+        let base_path = format!("{}/metadata_overwrite_replaces_existing", PATH);
         let _ = fs::remove_dir_all(&base_path);
         fs::create_dir_all(&base_path)?;
 
@@ -145,7 +144,7 @@ mod tests {
 
         let meta_back = Metadata::read_or_create(&base_path)?;
         assert_eq!(meta_back.last_purged_log_id, Some(vec![9]));
-        assert!(!fs::exists(format!("{base_path}/meta.hql.tmp")).unwrap_or(true));
+        assert!(!fs::exists(format!("{base_path}/meta.hql~")).unwrap_or(true));
 
         Ok(())
     }
--- a/hiqlite-wal/src/metadata.rs
+++ b/hiqlite-wal/src/metadata.rs
@@ -70,23 +70,32 @@ impl Metadata {
     #[inline]
     fn write_unchecked(bytes: &[u8], base_path: &str) -> Result<(), Error> {
         let path = format!("{base_path}/meta.hql");
+        let tmp_path = format!("{path}.tmp");
 
-        let _ = fs::remove_file(&path);
-        let mut file = File::create_new(&path)?;
-        // TODO overwriting the file when we started with .seek() did not work when the
-        // new meta was smaller than the old one, and therefore the CRC would not match.
-        // let mut file = OpenOptions::new()
-        //     .read(true)
-        //     .write(true)
-        //     .create(true)
-        //     .open(&path)?;
+        let mut file = File::create(&tmp_path)?;
 
         debug_assert_eq!(MAGIC_NO_META.len(), 7);
         file.write_all(MAGIC_NO_META)?;
         file.write_all(&[1u8])?;
         file.write_all(crc!(bytes).as_slice())?;
         file.write_all(bytes)?;
-        file.flush()?;
+        // make the data durable *before* the rename, so the rename cannot publish a
+        // partially flushed file
+        file.sync_all()?;
+        drop(file);
+
+        // atomic replace on the same filesystem: readers and crash recovery never observe a
+        // truncated or half-written metadata file (the old remove + create_new approach could
+        // leave the file missing entirely if we crashed in between)
+        #[cfg(not(unix))]
+        let _ = fs::remove_file(&path);
+        fs::rename(&tmp_path, &path)?;
+
+        // best-effort: make the rename itself durable across power loss
+        #[cfg(unix)]
+        if let Ok(dir) = File::open(base_path) {
+            let _ = dir.sync_all();
+        }
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


hiqlite-wal/src/metadata.rs
```diff
@@ -70,23 +70,30 @@ impl Metadata {
...
-        file.flush()?;
+        // make the data durable *before* the rename, so the rename cannot publish a
+        // partially flushed file
+        file.sync_all()?;
+        drop(file);
+
+        // atomic replace: readers/crash recovery never see a torn or missing meta file
+        #[cfg(not(unix))]
+        let _ = fs::remove_file(&path);
+        fs::rename(&tmp_path, &path)?;
```
sebadob, line 89: Like it is not, it's not atomic as the comment states. There are 2 steps:
- removing the file
- then doing the rename

When writing to a tmp file first, only doing `fs::rename` is atomic in the sance that it will replace any possibly existing original file. So to make this even more robust, don't to the `remove_file()` manually.

hiqlite-wal/src/metadata.rs
```diff
@@ -70,23 +70,30 @@ impl Metadata {
     #[inline]
     fn write_unchecked(bytes: &[u8], base_path: &str) -> Result<(), Error> {
         let path = format!("{base_path}/meta.hql");
+        let tmp_path = format!("{path}.tmp");
```
sebadob: I usually stick to the way most editors on linux do it and use `"{path}~"`. Usually, I would never request another commit just for that, but since we need a change for the write below anyway, it's no big deal.

hiqlite-wal/src/metadata.rs
```diff
@@ -117,4 +124,29 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn metadata_overwrite_is_atomic() -> Result<(), Error> {
```
sebadob: Does not really test for an atomic overwrite, which is actually not that easy to do. Not sure how it would behave if we can open the file manually and then trigger a write, if we can check it this way somehow.

I mean, the test seems to succeed right now, but the write it not atomic (manual remove upfront), so I would probably rename the test.

hiqlite/src/store/state_machine/memory/state_machine.rs
```diff
@@ -360,28 +360,39 @@ impl StateMachineMemory {
         let path = format!("{}/{}", self.path_snapshots, meta.snapshot_id);
         let path_temp = format!("{path}.temp");
```
sebadob: Does not belong to your changes, but I just saw that I used `.temp` here. Obviously, this line is very old. Would be nice if you use the same scheme as above with the appended `~` only, so we can be consistent.

hiqlite/src/store/state_machine/memory/state_machine.rs
```diff
@@ -360,28 +360,39 @@ impl StateMachineMemory {
...
         {
-            let mut file = fs::File::create_new(&path_temp)
+            // truncate any leftover temp from a crashed previous run
+            let mut file = fs::File::create(&path_temp)
                 .await
                 .map_err(|err| StorageIOError::write_state_machine(&err))?;
             file.write_all(snapshot_bytes)
                 .await
                 .map_err(|err| StorageIOError::write_state_machine(&err))?;
+            file.sync_all()
```
sebadob: This `sync_all()` is not strictly necessary. The reason why it's wrapped in `{}` is that the file is dropped directly, and the `write_all()` makes sure it's safe at least inside the kernels buffers. The kernel will also guarantee correctness for the following `fs::copy()` even if it's not synced to disk yet. This means we would lose the snapshot in case of a power outage, but not if the process crashes. However, losing a snapshot in that case is recovered with the next restart automatically, since a snapshot is created before any logs are deleted. This means we can always safely restore.

But, of course syncing is "more safe" and in general a good practice, you're right. I would change from `sync_all()` to `sync_data()` though, which is faster and good enough. We only care about the content. While looking at it, an `fs::rename()` would also be better than an `fs::copy()` with later manual delete.

hiqlite/src/store/state_machine/memory/state_machine.rs
```diff
@@ -360,28 +360,39 @@ impl StateMachineMemory {
...
+        fs::rename(&path_temp, &path)
             .await
             .map_err(|err| StorageIOError::write_state_machine(&err))?;
 
         // cleanup task for old snapshots
         let id = meta.snapshot_id.clone();
         let dir = self.path_snapshots.clone();
         task::spawn(async move {
-            let mut entries = fs::read_dir(&dir).await.unwrap();
+            let Ok(mut entries) = fs::read_dir(&dir).await else {
```
sebadob: Why did you change to a checked read here? Have you ever triggered a `panic` because of the `unwrap` above? Because it should be impossible that this ever happens. We just worked inside this dir. Whenever I `unwrap()`, I do this to spot erorr that should never happen anyway quickly instead of silently losing them. If we ever had a logic error here and read the wrong dir, we would never even know about it, because of the (somewhat) silent `return`. It's easy to miss a log entry.

hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -738,9 +738,17 @@ Got:      {}
 }
 
 #[inline]
-fn create_snapshot(conn: &rusqlite::Connection, path: String) -> Result<(), rusqlite::Error> {
-    let q = format!("VACUUM main INTO '{path}'");
-    conn.execute(&q, ())?;
+fn create_snapshot(conn: &rusqlite::Connection, path: String) -> Result<(), Error> {
+    // vacuum into a temp file and move it into place, so a crash can never leave a
+    // partially written snapshot at the final path
+    let path_temp = format!("{path}.temp");
```
sebadob: You be nice to use the already mentioned theme with the `~` instead of `.tmp` for consistency. I guess I did the old `.temp` or `.tmp` in a few other places as well, which would need an update in the future.

hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -761,7 +769,19 @@ fn create_backup(
     let path_full = format!("{target_folder}/{file}");
     info!("Creating database backup into {path_full}");
 
-    conn.execute(&format!("VACUUM main INTO '{path_full}'"), ())?;
+    // vacuum into a temp file and move it into place, so a crash mid-backup can never leave
+    // a partial file under the final backup name (restore would pick it up as valid)
+    let path_temp = format!("{path_full}.temp");
```
sebadob: Same as above, please use `~` instead of `.temp`.

hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -777,12 +797,28 @@ fn create_backup(
...
+                match s3.push(&path_full, &file).await {
+                    Ok(_) => {
+                        info!("Push backup to S3 has been finished");
+                        break;
+                    }
+                    Err(err) if attempt < 3 => {
+                        error!(
+                            "Error pushing Backup to S3 (attempt {attempt}/3): {err} - retrying"
+                        );
+                        time::sleep(Duration::from_secs(5 * attempt)).await;
```
sebadob, line 816: I think a higher upper limit like `10` would be good. Maybe use a `const S3_MAX_RETRIES` for this so we can adapt more easily in the future. We could then also re-use the same value in other places like backup restore in the future.

sebadob (changes requested):

reneleonhardt:
All comments addressed with full compliance — three follow-up commits, plus a few necessary changes on top.

- Memory snapshot restore keeps skipping legacy `.temp` leftovers in addition to `~`, so an upgrade can never pick up an unfinished snapshot written by an older version. Covered by a new unit test (`read_current_snapshot_skips_temp_files`).
- The cleanup task's `remove_file` errors stay warn-on-error rather than `unwrap`: two cleanup tasks can race on the same old snapshot, and a panic there would take down one of them for no reason. Only the `read_dir` is back to `unwrap`.
- `snapshot_builder.rs` keeps `.temp` for now, per your note that other spots can follow later.
- One heads-up on the metadata rename: `std::fs::rename` only replaces an existing file on POSIX — on Windows the overwrite would now error. Since the remove-then-rename window was exactly what you asked to close, I went with the plain rename; happy to add a Windows-specific replace fallback if you want one.

sebadob (approved):

sebadob:
> One heads-up on the metadata rename: std::fs::rename only replaces an existing file on POSIX — on Windows the overwrite would now error. Since the remove-then-rename window was exactly what you asked to close, I went with the plain rename; happy to add a Windows-specific replace fallback if you want one.

Are you sure it does not work on Windows? The tokio docs mention this behavior, and after a very quick lock at the source, they do have different branches depending on the OS, so I would assume that tokio handles this for us already. Anyway, I actually don't really care about Windows, but if this would ever become an issue, we could add specific handling.

Thanks.

reneleonhardt:
You're welcome!

I mean, you didn't merge my amazing new super fast CI in 360 yet, we could just add windows and macos if you want to expand coverage 😉
No dev has Windows or free devcontainers, I couldn't double check 🤷

sebadob:
> I mean, you didn't merge my amazing new super fast CI in 360 yet

It has some failing tests, that's why. Did not have the time yet to look into it, about what exactly it is.

> No dev has Windows or free devcontainers, I couldn't double check

Me neither. I don't even own an old PC anymore that still runs it, so I don't really care too much.

PR #358, opened 2026-08-12
fix: Cap rate-limit buckets, surface truncated reads, and make migrations deterministic

reneleonhardt:
## Summary

A set of small correctness fixes across the client, query, and backup paths:

- The rate-limit token bucket refilled with `max(current + rps, burst)`, so the counter grew without bound after any idle period and the configured burst limit became meaningless. Refill now caps at `burst`.
- Every read path silently stopped on a mid-query row error and returned `Ok(partial_rows)` — `query_map_one` could even return a "correct" single row from truncated data. Read errors now propagate.
- Migration timestamps used the wall clock inside the raft state machine, so `_migrations.ts` diverged between nodes. The timestamp now comes from the raft log id.
- The WAL purged-log metadata was persisted after log files were deleted; a crash in between left a stale frontier pointing at deleted files.
- SQLite connections had no busy timeout, so reads failed immediately with `SQLITE_BUSY` during backup and snapshot-restore windows.
- Directory-listing loops swallowed `read_dir` errors silently, which could select a stale snapshot; they now warn.
- Client-side migration mismatches and backup-task panics crashed the calling task or silently stopped backups; they now return errors.

## Changes

- Rate limiter: `min(current + rps, burst)`, saturating.
- Query + dashboard read paths: `loop`/`match` over `rows.next()`, errors returned instead of truncating.
- `VecText` mapped from a non-text column returns a type error instead of silently becoming an empty value.
- Migration: `apply_migration` takes the raft log id as the deterministic timestamp.
- WAL writer: `last_purged_log_id` persisted before log files are deleted.
- SQLite connections: 5 s busy timeout.
- Snapshot/backup dir listings: `warn!` on read errors.
- `client.migrate()`: id/hash mismatches return `Err` instead of panicking.
- Backup cron and snapshot-trigger failures log and continue instead of panicking the backup task.

## Verification

- `cargo test -p hiqlite --lib --features full` — 24 passed (branch base; re-verify on rebase).
- `cargo test -p hiqlite-wal` — 7 passed.

## Compatibility notes

The busy timeout changes read behavior from immediate `SQLITE_BUSY` errors to waiting up to 5 s during lock-held windows. The migration `ts` column value changes (raft log id instead of wall clock) — nothing reads it semantically.

### DIFF
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -307,10 +307,8 @@ fn run(
                     is_dirty = false;
                 }
 
-                // Persist `last_purged_log_id` before deleting WAL files: a stale (too low)
-                // value would point into deleted files; a too-high one keeps extra files (safe).
-                // If the deletion fails, the metadata is reverted again below, so it never
-                // claims logs as purged that are still present on disk.
+                // Persist the purge frontier before deleting (too low = hole into deleted files;
+                // too high = extra files). Revert it again if the deletion fails below.
                 let previous_purged = meta.read()?.last_purged_log_id.clone();
                 let persist_purged = last_log.is_some();
                 if persist_purged {
@@ -331,12 +329,9 @@ fn run(
                     }
                     Err(err) => {
                         if persist_purged {
-                            // revert the persisted frontier: the logs were not deleted, so the
-                            // metadata must not claim them as purged. Even if the revert write
-                            // fails, the original deletion error is still reported to the caller.
+                            // deletion failed: revert the frontier, but still report the error
                             let revert_res = {
-                                // release the write lock before persisting, since
-                                // `Metadata::write` acquires a read lock on the same meta
+                                // `Metadata::write` re-locks the meta, so release the guard first
                                 let mut m = match meta.write() {
                                     Ok(m) => m,
                                     Err(poisoned) => poisoned.into_inner(),
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -307,15 +307,19 @@ fn run(
                     is_dirty = false;
                 }
 
+                // Persist `last_purged_log_id` BEFORE deleting any WAL files. If we crashed in
+                // between, we would end up with a `last_purged_log_id` pointing into already
+                // deleted files, which breaks recovery via `get_log_state()`. A too-low value on
+                // the other hand only keeps some extra files around, which is safe.
+                if last_log.is_some() {
+                    meta.write()?.last_purged_log_id = last_log;
+                    Metadata::write(meta.clone(), &wal.base_path)?;
+                }
+
                 buf.clear();
                 buf_logs.clear();
                 match wal.shift_delete_logs(from, until, wal_size, &mut buf, &mut buf_logs) {
                     Ok(_) => {
-                        // the last_log may be none if logs are truncated
-                        if last_log.is_some() {
-                            meta.write()?.last_purged_log_id = last_log;
-                            Metadata::write(meta.clone(), &wal.base_path)?;
-                        }
                         {
                             let mut lock = wal_locked.write().unwrap();
                             lock.active = wal.active;
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -307,8 +307,8 @@ fn run(
                     is_dirty = false;
                 }
 
-                // Persist `last_purged_log_id` before deleting WAL files: a stale (too high)
-                // value would point into deleted files; a too-low one keeps extra files (safe).
+                // Persist `last_purged_log_id` before deleting WAL files: a stale (too low)
+                // value would point into deleted files; a too-high one keeps extra files (safe).
                 // If the deletion fails, the metadata is reverted again below, so it never
                 // claims logs as purged that are still present on disk.
                 let previous_purged = meta.read()?.last_purged_log_id.clone();
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -307,10 +307,8 @@ fn run(
                     is_dirty = false;
                 }
 
-                // Persist `last_purged_log_id` BEFORE deleting any WAL files. If we crashed in
-                // between, we would end up with a `last_purged_log_id` pointing into already
-                // deleted files, which breaks recovery via `get_log_state()`. A too-low value on
-                // the other hand only keeps some extra files around, which is safe.
+                // Persist `last_purged_log_id` before deleting WAL files: a stale (too high)
+                // value would point into deleted files; a too-low one keeps extra files (safe).
                 if last_log.is_some() {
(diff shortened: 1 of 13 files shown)


hiqlite-wal/src/writer.rs
```diff
@@ -307,15 +307,17 @@ fn run(
                     is_dirty = false;
                 }
 
+                // Persist `last_purged_log_id` before deleting WAL files: a stale (too high)
+                // value would point into deleted files; a too-low one keeps extra files (safe).
+                if last_log.is_some() {
+                    meta.write()?.last_purged_log_id = last_log;
+                    Metadata::write(meta.clone(), &wal.base_path)?;
+                }
```
sebadob, line 317: While this makes sense to do before the deletion, you need to revert it if the deletion fails. A failure will (usually) happen when the file cannot be removed from disk. That mmapping fails at runtime if very unlikely.

hiqlite/src/client/migrate.rs
```diff
@@ -46,17 +46,25 @@ impl Client {
                 }
                 Some(to_migrate) => {
                     if to_migrate.id != migration.id {
-                        panic!(
-                            "ID mismatch for '{}' between given and already applied migration: {} != {}",
-                            to_migrate.name, to_migrate.id, migration.id
-                        );
+                        // this is a runtime condition depending on the server DB state, so it
+                        // must be an error, not a panic in the caller's task
```
sebadob: While migrations will be triggered at runtime, this is not a runtime condition. Migrations are read from files which are added at compile time, or actually even beforehand. A `panic` is much preferred, because the DB will be in an inconsistent state if that ever happens. The same is true for the `panic` below.

Running on an inconsistent DB is the worst thing you can do, and you should never proceed in such a situation. This always(!) requires user interaction to resolve.

hiqlite/src/client/rate_limit.rs
```diff
@@ -37,11 +37,11 @@ impl Client {
             interval.tick().await;
 
             #[cfg(feature = "cache")]
-            if let Some(config) = &config_cache {
-                let lim = slf.inner.rate_limit_cache.as_ref().unwrap();
-
+            if let Some(config) = &config_cache
+                && let Some(lim) = slf.inner.rate_limit_cache.as_ref()
```
sebadob: The check here for the `slf.inner.rate_limit_cache` instead of the above `unwrap()` is unnecessary. It only costs (a tiny bit) of performance. If the `unwrap()` ever triggers, it's a logic bug during client initialization. This will show up immediately in integration tests. 

Doing this check is even much worse as it will lead to a silent failure that must never happen anyway, and this would be terrible to debug. When I `panic`, `unwrap` or `expect` anywhere, I am not doing it out of lazyness, but for a good reason, usually to catch bugs that must never happen anyway fast.

hiqlite/src/client/rate_limit.rs
```diff
@@ -124,3 +124,27 @@ impl Client {
         }
     }
 }
+
+/// Token bucket refill: add `rps` tokens every second, never exceeding `burst`.
+#[inline]
+fn refill_bucket(current: u32, rps: u32, burst: u32) -> u32 {
+    min(current.saturating_add(rps), burst)
```
sebadob, line 132: The `max()` logic bug here is a big one, thanks! A bit embarassing that it was a `max` before.

hiqlite/src/store/state_machine/sqlite/state_machine.rs
```diff
@@ -382,6 +382,10 @@ impl StateMachineSqlite {
         conn.pragma_update(None, "auto_vacuum", "INCREMENTAL")?;
         conn.pragma_update(None, "optimize", "0x10002")?;
 
+        // Backups/snapshot restores hold an exclusive lock; the busy timeout stops concurrent
+        // reads from failing with `SQLITE_BUSY` during those windows.
+        conn.busy_timeout(Duration::from_secs(5))?;
```
sebadob: I did not add it because it should never happen when in WAL mode, since there is only a single writer connection allowed, but I totally forgot about backups / VACUUM. Good catch. 

Though, I would even increase the timeout to at least 10 seconds. This should make everything work for even a bit bigger databases. Maybe even to 30, since a backup is the only situation when this should ever be necessary.

hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -791,7 +791,17 @@ fn create_backup(
...
+fn migrate(
+    conn: &mut rusqlite::Connection,
+    mut migrations: Vec<Migration>,
+    last_applied_log_id: Option<LogId<NodeId>>,
+) -> Result<(), Error> {
+    // The `ts` written to `_migrations` must be identical on every node, because the migration
+    // entry is applied by the Raft state machine on all members. Using the wall clock here would
+    // make the nodes diverge, which is exactly what the non-deterministic function overrides in
+    // `overwrite_non_det_fns` are meant to prevent. The Raft log id is deterministic and unique.
+    let migration_ts = last_applied_log_id.map(|l| l.index as i64).unwrap_or(0);
```
sebadob, line 801: Just as a side note, while this is true, that it would be inconsistent, the migration TS is only ever meant if you need to manually debug something on a node. This is not used in any other context than just being informational. You could even argue that you want to have the wall block to identify things like slow nodes, but that's not important. This solution now is the cleaner one for sure.

hiqlite/src/backup.rs
```diff
@@ -201,7 +203,18 @@ pub(crate) async fn backup_local_cleanup(backup_path: String, keep_days: u16) ->
...
-    while let Ok(Some(entry)) = dir_entries.next_entry().await {
+    loop {
+        let entry = match dir_entries.next_entry().await {
+            Ok(Some(entry)) => entry,
+
+            Ok(None) => break,
+
+            Err(err) => {
+                warn!("Error reading directory entries: {err:?}");
+
```
sebadob: These newlines in between look a bit unwanted I guess?

sebadob (changes requested):

reneleonhardt:
All review comments addressed.

- **WAL purge frontier:** persisted before log deletion (crash-window fix kept), and **reverted if the deletion fails** so metadata never claims logs as purged that are still on disk; the original deletion error is always reported to the caller.
- **Migration mismatches:** id/hash mismatches `panic!` again — migrations are compile-time embedded, so a mismatch means an inconsistent DB that must never run.
- **Rate-limit ticker:** `unwrap()` restored — config and limiter are set together, so a missing limiter is an init bug that fails loudly instead of silently skipping refills.
- **Busy timeout:** raised from 5 s to 30 s for bigger backup/snapshot-restore windows.
- **backup.rs:** stray blank lines removed from the dir-listing `match`.
- Comments 4 and 6: thanks; the raft-log-id migration `ts` stays as the cleaner, deterministic solution.
- Comments minimized on the branch — only short "why" comments kept where behavior is non-obvious.

sebadob (approved):
Thanks

PR #359, opened 2026-08-12
fix(sqlite): Keep the writer thread alive on user input

reneleonhardt:
## Summary

The SQLite state machine runs on a single writer thread. Several user-reachable inputs could panic inside it, killing the thread and with it the whole database until a node restart (the thread is unsupervized, so the node became a zombie until a health check restarted it):

- Non-deterministic SQL functions (`now()`, `date()`, `random()`, `strftime()` and friends) were registered on the write connection to `panic!` — one `INSERT ... now()` from any application killed the node.
- Migration validation failures (gaps, missing or changed migrations) panicked.
- Snapshot restore failures and corrupt restored metadata panicked.
- A dropped client response channel panicked the writer (`expect` on a oneshot send), so a cancelled or timed-out request could take down the database.
- `Param::StmtOutput*` values reaching a plain `execute`/`execute_returning` panicked during binding.
- A column-read failure while building the observable first row of a transaction statement panicked.

All of these now return regular errors; the writer thread survives every one of them.

## Changes

- Non-deterministic function guards return a `UserFunctionError` instead of panicking (only the offending statement fails).
- Migration validation returns `Error` instead of panicking.
- Snapshot restore errors and metadata-read failures propagate through the ack channel instead of panicking.
- All response-channel sends tolerate dropped receivers (`.ok()` / graceful handling) instead of `expect`/`unwrap` panicking.
- `Param::into_sql` returns `Result`; the writer bind loops surface the error.
- Transaction first-row column reads produce a query error + rollback instead of a panic.

## Verification

- `cargo test -p hiqlite --lib --features full` — 25 passed (branch base; re-verify on rebase), including tests that `SELECT now()` errors while the connection stays usable, that migration gaps return errors, and that `StmtOutput` params error outside transactions.

## Compatibility notes

Applications that relied on the panic as a hard "determinism violation" signal now get a SQL error instead; the statement fails identically on every member, so the raft state stays consistent.

### DIFF
--- a/hiqlite/src/store/state_machine/sqlite/param.rs
+++ b/hiqlite/src/store/state_machine/sqlite/param.rs
@@ -32,7 +32,7 @@ pub enum Param {
 }
 
 impl Param {
-    pub(crate) fn into_sql<'a>(self) -> ToSqlOutput<'a> {
+    pub(crate) fn into_sql<'a>(self) -> Result<ToSqlOutput<'a>, Error> {
         let value = match self {
             Param::Null => Value::Null,
             Param::Integer(i) => Value::Integer(i),
@@ -40,10 +40,15 @@ impl Param {
             Param::Text(t) => Value::Text(t),
             Param::Blob(b) => Value::Blob(b),
             Param::StmtOutputNamed(..) | Param::StmtOutputIndexed(..) => {
-                panic!("Param::StmtOutput is only valid inside transactions")
+                // StmtOutput is resolved inside transactions only. If it reaches the writer
+                // thread via a plain execute/execute_returning, panicking would kill the writer
+                // and with it the whole database -> return a regular error instead.
+                return Err(Error::QueryParams(
+                    "Param::StmtOutput is only valid inside transactions".into(),
+                ));
             }
         };
-        ToSqlOutput::Owned(value)
+        Ok(ToSqlOutput::Owned(value))
     }
 
     pub(crate) fn into_sql_txn_ctx<'a>(
@@ -294,6 +299,17 @@ impl From<StmtColumn<String>> for Param {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn stmt_output_param_errors_outside_transactions() {
+        let err = Param::StmtOutputNamed(0, "col".into())
+            .into_sql()
+            .unwrap_err();
+        assert!(err.to_string().contains("only valid inside transactions"));
+
+        let err = Param::StmtOutputIndexed(0, 0).into_sql().unwrap_err();
+        assert!(err.to_string().contains("only valid inside transactions"));
+    }
     use crate::query::rows::ValueOwned;
     use uuid::{NoContext, Timestamp};
 
--- a/hiqlite/src/store/state_machine/sqlite/param.rs
+++ b/hiqlite/src/store/state_machine/sqlite/param.rs
@@ -4,7 +4,6 @@ use super::{
     transaction_env::{TransactionEnv, TransactionParamContext},
     transaction_variable::StmtColumn,
 };
-use crate::Error;
 use chrono::{DateTime, FixedOffset, Local, NaiveDate, NaiveDateTime, NaiveTime, Utc};
 use rusqlite::types::{ToSqlOutput, Value};
 use serde::{Deserialize, Serialize};
@@ -32,7 +31,7 @@ pub enum Param {
 }
 
 impl Param {
-    pub(crate) fn into_sql<'a>(self) -> Result<ToSqlOutput<'a>, Error> {
+    pub(crate) fn into_sql<'a>(self) -> ToSqlOutput<'a> {
         let value = match self {
             Param::Null => Value::Null,
             Param::Integer(i) => Value::Integer(i),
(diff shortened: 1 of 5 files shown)


hiqlite/src/store/state_machine/sqlite/param.rs
```diff
@@ -32,18 +32,23 @@ pub enum Param {
...
             Param::Integer(i) => Value::Integer(i),
             Param::Real(r) => Value::Real(r),
             Param::Text(t) => Value::Text(t),
             Param::Blob(b) => Value::Blob(b),
             Param::StmtOutputNamed(..) | Param::StmtOutputIndexed(..) => {
-                panic!("Param::StmtOutput is only valid inside transactions")
+                // StmtOutput is resolved inside transactions only. If it reaches the writer
+                // thread via a plain execute/execute_returning, panicking would kill the writer
+                // and with it the whole database -> return a regular error instead.
+                return Err(Error::QueryParams(
```
sebadob: Well, this (the same as in other PRs), was a `panic` on purpose. Removing all `panic`s is not a goal or even desirable.  You must never use this param outside of a transaction. This must be `panic`. A `Result<_>` only makes sense for recoverable error. This is non-recoverable, it needs a code logic update, and there is nothing louder than a `panic`.

hiqlite/src/store/state_machine/sqlite/state_machine.rs
```diff
@@ -399,116 +399,89 @@ impl StateMachineSqlite {
     }
 
     fn overwrite_non_det_fns(conn: &rusqlite::Connection) {
+        // Non-deterministic functions are forbidden on raft write connections (nodes must
+        // apply identical statements); return an error instead of panicking the writer.
```
sebadob: In all of these cases, again, the `panic` is 100% on purpose. All of them are non-recoverable, except for the dashboard. But the dashboard is for debugging purposes / manual stuff. If you actually want to prevent that from a panic, there is a better solution without degrading DX by returning (somewhat silent) errors here. You should scan the query strings for manual dashboard queries in advance and show a proper error message for best UX.

hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -215,14 +226,14 @@ CREATE TABLE IF NOT EXISTS _metadata
                             }
 
                             if let Some(err) = params_err {
-                                q.tx.send(Err(err)).expect("oneshot tx to never be dropped");
```
sebadob, line 218: For these send's, the same as always, I `expect` on purpose to detect logic errors properly. If this ever happens, the whole state machine would be gone, which must not happen.

Did you actually produce a `panic` for these? If so, how and when? This should never happen. The question then is why did it happen and how to solve it, not silently dropping it.

hiqlite/src/store/state_machine/sqlite/writer.rs
```diff
@@ -556,16 +585,33 @@ CREATE TABLE IF NOT EXISTS _metadata
                         start.elapsed().as_millis()
                     );
 
-                    sm_data = conn
-                        .query_row("SELECT data FROM _metadata WHERE key = 'meta'", (), |row| {
-                            let meta_bytes: Vec<u8> = row.get(0)?;
-                            let metadata: StateMachineData =
-                                deserialize(&meta_bytes).expect("Metadata to deserialize ok");
```
sebadob, line 563: `expect` on purpose. Failure during metadata deserialization is non-recoverable. This is the case for lots of other lines in this file, or in general.

sebadob (changes requested):
I mentioned it already, but simply removing all `panic`, `unwrap` or even `expect` is neither a goal nor desirable. I do this on purpose to catch logic or non-recoverable errors quickly. An error in some logs is easily overseen, a `panic` is not, and you should never compromise something like DB consistency over "just keep it alive at all costs".

reneleonhardt:
Agreed, complying.

- Non-deterministic functions panic again on the write connection; the dashboard pre-scans manual queries and returns a proper error instead.
- `Param::StmtOutput*` outside a transaction panics again.
- Writer response-channel sends `expect` again.
- Snapshot-restore metadata read/deserialize expects again.

Other changes:

- Migration validation (gaps, missing or changed migrations) panics again — an inconsistent DB must never run.
- Startup metadata deserialization expects again.
- Restore failure stays a recoverable error openraft can retry; transaction first-row read failures stay query errors.
- Tests updated to assert the panic semantics; 35 lib tests pass, clippy clean.

sebadob (dismissed):
Thanks

sebadob:
Just needs a quick conflict resolve.

Apart from that, we could think about a feature like `non_det_fn_rewirte`. If set, instead of checking for non det fns and returning an error (or panic via programmatic use), we could rewrite them to make them absolute. This should not be the default, because checking each query is actually quite expensive to do. 

The only issue with that is that queries are being prepared and cached. This means if we rewrite these, and they are used consistently instead of doing it via `params!()`, the cache for prepared queries would not be finite. This would technically work, because old entries are being evicted, but it's a waste of resources that would otherwise easily solved when done correctly from the beginning. So, I am not sure about this.

I think the current approach rejecting non det fns for all modifying queries is the better choice.

reneleonhardt:
Thanks for the review — all four points are already addressed in the branch.
Rebased because of conflicts, optimized wording along the way.

- `Param::StmtOutput*` outside a transaction panics.
- The forbidden non-deterministic functions panic on the write connection (rusqlite turns that into a statement error, so the writer survives).
- Response-channel sends `expect`.
- Metadata read/deserialize after a snapshot restore expects.

Also: migration validation panics; the dashboard pre-scans manual write queries for the forbidden functions and returns a readable error.
Ready when you are.

sebadob (approved):
Thanks

PR #361, opened 2026-08-15
fix(wal): only tell openraft an append is durable after it is on disk

drmingdrmer:
Four places tell openraft that data is durable before it is. I found these while reviewing how projects use openraft — I maintain [openraft](https://github.com/databendlabs/openraft) — and `hiqlite-wal` is the only implementation I looked at that had the async-callback machinery in the right shape already, which is why the fix is small.

Happy to split this into separate PRs, or to drop the `interval_<ms>` part (see the last section) if you'd rather keep that mode as-is.

## 1. The append callback runs before the flush

```rust
if sync == LogSync::Immediate {
    wal.active().flush()?;
} else if sync == LogSync::ImmediateAsync {
    wal.active().flush_async()?;
} else {
    is_dirty = true;
}
// TODO with the next big openraft release, we can do async callbacks
callback();
```

That `callback` is openraft's `LogFlushed` notification. Running it asserts *"these entries are on disk"*, and the leader commits on the strength of that assertion — [`RaftLogStorage::append`](https://docs.rs/openraft/0.9/openraft/storage/trait.RaftLogStorage.html#tymethod.append): *"When the callback is called, the entries must be persisted on disk."*

Under `Immediate` that holds. Under `ImmediateAsync` (**the default**) `flush_async` only issues `msync(MS_ASYNC)` and returns; under `interval_<ms>` nothing has been flushed at all. In both cases the callback fires while the data is in the page cache, so a power loss or kernel panic drops entries that a quorum already reported durable — and the leader has already told the client the write succeeded.

**On the TODO:** async callbacks do not need a newer openraft, and never did. Openraft 0.9 already documents *"the `callback` can be called either before or after this method returns"* — that allowance exists precisely so a WAL implementation can do what this PR does.

### Fix: queue the callbacks, release them behind a real flush

```rust
fn flush_and_release(
    wal: &mut WalFileSet,
    buf: &mut Vec<u8>,
    is_dirty: &mut bool,
    pending: &mut Vec<Box<dyn FnOnce() + Send>>,
) -> Result<(), Error> {
    if *is_dirty {
        let active = wal.active();
        buf.clear();
        active.update_header(buf)?;
        active.flush()?;
        *is_dirty = false;
    }
    for callback in pending.drain(..) {
        callback();
    }
    Ok(())
}
```

Deferring is what makes **group commit** possible, so this costs much less than one fsync per append:

| mode | behaviour |
|---|---|
| `Immediate` | unchanged — flush, then release |
| `ImmediateAsync` | keep accumulating while `rx.is_empty()` is false, then one blocking flush releases the whole burst |
| `interval_<ms>` | released by `Action::Sync` on the next tick |

Under `ImmediateAsync`, a burst of appends now pays **one** fsync between them instead of one per append; an isolated append pays one, same as `Immediate`. So the default mode becomes genuinely durable, and stays fast exactly when it matters — under load.

One small behaviour change inside the helper: the `Immediate` append path now calls `update_header` before flushing, which it previously skipped. `Action::Sync` and `Action::Remove` always did, so this makes the three consistent.

## 2. `Action::Remove` flushed asynchronously before deleting logs

```rust
// Before removing any logs, make sure that all in-memory buffers are flushed. If
// at least headers and metadata are not up to date, and a crash happens in the
// middle of removing logs, we could end up with a hole between Snapshot and latest
// existing Raft Log, which must never happen.
active.flush_async()?;
```

The comment states the requirement correctly, but `flush_async` does not meet it: it starts the msync and returns, so a crash during the removal below can still land after the deletions and before the header reaches disk. That is the hole the comment is guarding against. Now uses the blocking helper.

## 3. `Action::Vote` cleared `is_dirty` after an async flush

```rust
wal.active().flush_async()?;
is_dirty = false;
```

The WAL is marked clean while the msync may still be in flight, so the interval ticker then skips a file that never reached disk. Also uses the blocking helper now.

## 4. `Metadata::write_unchecked` — the vote file

```rust
let _ = fs::remove_file(&path);
let mut file = File::create_new(&path)?;
...
file.flush()?;
```

`meta.hql` holds the raft **vote** and `last_purged_log_id`. Openraft requires the vote to be on disk before `save_vote` returns, because that is what makes "a node votes at most once per term" hold. Two problems:

- **The delete-then-create window.** Between `remove_file` and a completed write, the file does not exist. A crash there loses the vote outright, not just the newest value.
- **`File::flush()` is a no-op.** For `std::fs::File`, `Write::flush` flushes Rust-side buffering, of which there is none. It never issues an fsync. `sync_all()` is the call that does.

Either way the node can come back with an older vote and grant its vote a second time in the same term — two candidates each collect a majority, and both are leader for that term.

Now: write `meta.hql.tmp`, `sync_all()`, rename over `meta.hql`, then fsync the directory (a rename is itself a directory update — without that last step the renamed file can be missing after a crash even though its contents were flushed). `sync_dir` is `#[cfg(unix)]`; on Windows it is a no-op.

Dropping the delete-then-create also removes the reason for the `TODO` comment above it: the seek-based overwrite failed when the new metadata was shorter than the old, and a fresh temp file plus rename has no such problem.

## Behaviour change to decide on: `interval_<ms>`

Deferring the callback to the ticker means an openraft commit under `interval_200` can now take up to 200 ms. That is what the setting asks for — flushes happen every 200 ms, so nothing can be durable sooner — but it is a latency change for anyone running it today, and it makes the mode much less attractive than the docs currently suggest.

Two options, your call:

1. **As in this PR.** `interval_<ms>` becomes honest: slower to commit, never loses acknowledged entries.
2. **Revert just that arm** and keep firing the callback immediately for `interval_<ms>` only, documenting it as explicitly non-durable. The `ImmediateAsync` fix — which covers the default and is a strict improvement with no latency cost under load — stands on its own.

Worth noting for the docs either way: rauthy's tuning page quotes this config block verbatim — caution included — with `log_sync = "interval_200"` as the shown value, and nothing on the page says that value assumes a multi-node cluster. A single-instance deployment has no other replica to recover the missing entries from, which is exactly what the caution warns about.

## Verification

```
cargo check -p hiqlite-wal
cargo test  -p hiqlite-wal                                  # 8 passed, 0 failed
cargo check -p hiqlite --features server,sqlite,cache
cargo test  -p hiqlite --features server,sqlite,cache --lib # 38 passed, 0 failed
```

Durability cannot be asserted from inside the process — a test would have to kill the machine, not the process — so the change is instead scoped so that every flush site goes through one helper.

### DIFF
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -142,6 +142,28 @@ pub fn spawn(
     Ok((tx, wal_locked))
 }
 
+/// Flush the active WAL file so that everything written to it is on disk.
+///
+/// `flush_async` only starts the msync and returns, so a WAL flushed that way is
+/// not known to be on disk. Only a flush that returns may clear `is_dirty`.
+fn flush_blocking(
+    wal: &mut WalFileSet,
+    buf: &mut Vec<u8>,
+    is_dirty: &mut bool,
+) -> Result<(), Error> {
+    if !*is_dirty {
+        return Ok(());
+    }
+
+    let active = wal.active();
+    buf.clear();
+    active.update_header(buf)?;
+    active.flush()?;
+    *is_dirty = false;
+
+    Ok(())
+}
+
 fn spawn_syncer(tx_writer: flume::Sender<Action>, mut interval: Interval) {
     task::spawn(async move {
         loop {
@@ -257,12 +279,14 @@ fn run(
                     error!("error sending back ack after logs append: {err:?}");
                 }
 
+                // The WAL now holds bytes that are not known to be on disk, and only a
+                // blocking flush can clear that state again. `flush_async` merely starts the
+                // msync, which is why `Action::Remove` and `Action::Vote` below still flush.
+                is_dirty = true;
                 if sync == LogSync::Immediate {
-                    wal.active().flush()?;
+                    flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
                 } else if sync == LogSync::ImmediateAsync {
                     wal.active().flush_async()?;
-                } else {
-                    is_dirty = true;
                 }
                 // TODO with the next big openraft release, we can do async callbacks
                 callback();
@@ -297,15 +321,11 @@ fn run(
                 // at least headers and metadata are not up to date, and a crash happens in the
                 // middle of removing logs, we could end up with a hole between Snapshot and latest
                 // existing Raft Log, which must never happen.
-                let active = wal.active();
-                if is_dirty {
-                    buf.clear();
-                    active.update_header(&mut buf)?;
-                    // async is fine, as long as we trigger it before starting log removal.
-                    // If the flush fails, so would the log removal, and we would not have a hole.
-                    active.flush_async()?;
-                    is_dirty = false;
-                }
+                //
+                // The flush has to block. `flush_async` only starts the msync and returns, so a
+                // crash during the removal below can still land after the deletions and before
+                // the header reaches disk, which is the hole this guards against.
+                flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
 
                 // Persist the purge frontier before deleting (too low = hole into deleted files;
                 // too high = extra files). Revert it again if the deletion fails below.
@@ -353,9 +373,9 @@ fn run(
             Action::Vote { value, ack } => {
                 debug!("WAL Writer - Action::Vote");
 
-                buf.clear();
-                wal.active().flush_async()?;
-                is_dirty = false;
+                // Blocking on purpose: clearing `is_dirty` while an msync is still in flight
+                // would make the interval ticker skip a WAL that never reached disk.
+                flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
 
                 meta.write()?.vote = Some(value);
                 let res = Metadata::write(meta.clone(), &wal.base_path);
@@ -363,13 +383,9 @@ fn run(
                 ack.send(res).unwrap();
             }
             Action::Sync => {
-                if is_dirty {
-                    let active = wal.active();
-                    buf.clear();
-                    active.update_header(&mut buf)?;
-                    active.flush_async()?;
-                    is_dirty = false;
-                }
+                // The ticker is the only flush in `IntervalMillis` mode and no append waits on
+                // it, so it blocks: `msync(MS_ASYNC)` starts no writeback on Linux at all.
+                flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
             }
             Action::Shutdown(ack) => {
                 debug!("Raft logs store writer is being shut down");
--- a/hiqlite-wal/src/writer.rs
+++ b/hiqlite-wal/src/writer.rs
@@ -144,8 +144,8 @@ pub fn spawn(
 
 /// Flush the active WAL file so that everything written to it is on disk.
 ///
-/// `flush_async` only starts the msync and returns, so a WAL flushed that way is
-/// not known to be on disk. Only a flush that returns may clear `is_dirty`.
+/// `flush_async` only starts the writeback and returns, so a WAL flushed that way
+/// is not known to be on disk. Only a flush that returns may clear `is_dirty`.
 fn flush_blocking(
     wal: &mut WalFileSet,
     buf: &mut Vec<u8>,
@@ -281,14 +281,16 @@ fn run(
 
                 // The WAL now holds bytes that are not known to be on disk, and only a
                 // blocking flush can clear that state again. `flush_async` merely starts the
-                // msync, which is why `Action::Remove` and `Action::Vote` below still flush.
+                // writeback, which is why `Action::Remove` and `Action::Vote` below still flush.
                 is_dirty = true;
                 if sync == LogSync::Immediate {
                     flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
                 } else if sync == LogSync::ImmediateAsync {
                     wal.active().flush_async()?;
                 }
-                // TODO with the next big openraft release, we can do async callbacks
+                // openraft takes this callback as "these entries are on disk" and commits on
+                // a quorum of such acks. Only `Immediate` upholds that here: the async levels
+                // deliberately ack first and trade the writeback window for throughput.
                 callback();
 
                 // Roll WAL pre-emptively if only very few space is left at this point, because
@@ -322,9 +324,9 @@ fn run(
                 // middle of removing logs, we could end up with a hole between Snapshot and latest
                 // existing Raft Log, which must never happen.
                 //
-                // The flush has to block. `flush_async` only starts the msync and returns, so a
-                // crash during the removal below can still land after the deletions and before
-                // the header reaches disk, which is the hole this guards against.
+                // The flush has to block. `flush_async` only starts the writeback and returns,
+                // so a crash during the removal below can still land after the deletions and
+                // before the header reaches disk, which is the hole this guards against.
                 flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
 
                 // Persist the purge frontier before deleting (too low = hole into deleted files;
--- a/hiqlite-wal/Cargo.toml
+++ b/hiqlite-wal/Cargo.toml
@@ -36,3 +36,6 @@ thiserror.workspace = true
 thread-priority.workspace = true
 tokio.workspace = true
 tracing.workspace = true
+
+[target.'cfg(target_os = "linux")'.dependencies]
+libc = "0.2"
--- a/hiqlite-wal/src/wal.rs
+++ b/hiqlite-wal/src/wal.rs
@@ -42,6 +42,10 @@ pub struct WalFile {
     len_max: u32,
     mmap: Option<Mmap>,
     mmap_mut: Option<MmapMut>,
+    // Kept while `mmap_mut` exists: `sync_file_range` in `flush_async` needs an fd,
+    // which the mmap alone does not provide.
+    #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
+    file: Option<File>,
 }
 
 impl Drop for WalFile {
@@ -345,6 +349,7 @@ impl WalFile {
             len_max: self.len_max,
             mmap: None,
             mmap_mut: None,
+            file: None,
         }
     }
 
@@ -496,10 +501,32 @@ impl WalFile {
         Ok(())
     }
 
+    /// Starts writeback of all dirty pages without waiting for completion.
+    ///
+    /// On Linux this uses `sync_file_range(SYNC_FILE_RANGE_WRITE)`, because
+    /// `msync(MS_ASYNC)` starts no writeback there at all. This is not a durability
+    /// barrier: neither the device write cache nor the extent allocations of the
+    /// sparse WAL file are flushed, so data flushed this way must not be treated as
+    /// on disk and `is_dirty` must stay set.
     #[inline]
     pub fn flush_async(&mut self) -> Result<(), Error> {
         debug_assert!(self.mmap_mut.is_some());
+
+        #[cfg(target_os = "linux")]
+        {
+            use std::os::fd::AsRawFd;
+
+            let file = self.file.as_ref().expect("file kept while mmap_mut");
+            let res = unsafe {
+                libc::sync_file_range(file.as_raw_fd(), 0, 0, libc::SYNC_FILE_RANGE_WRITE)
+            };
+            if res != 0 {
+                return Err(std::io::Error::last_os_error().into());
+            }
+        }
+        #[cfg(not(target_os = "linux"))]
         self.mmap_mut.as_mut().unwrap().flush_async()?;
+
         Ok(())
     }
 
@@ -562,6 +589,7 @@ impl WalFile {
         mmap.advise(memmap2::Advice::Sequential)?;
 
         self.mmap_mut = Some(mmap);
+        self.file = Some(file);
 
         Ok(())
     }
@@ -596,6 +624,7 @@ impl WalFile {
             len_max: wal_size,
             mmap: None,
             mmap_mut: None,
+            file: None,
         })
     }
 
@@ -667,6 +696,7 @@ impl WalFile {
             len_max: len_max as u32,
             mmap: None,
             mmap_mut: None,
+            file: None,
         })
     }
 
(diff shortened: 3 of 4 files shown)


sebadob:
Don't worry about the CI, I aborted it. It's behaving like this sometimes, even though everything should be fine, not sure why. Will have a look at your PR later as well.

hiqlite-wal/src/metadata.rs
```diff
@@ -67,26 +67,45 @@ impl Metadata {
...
         file.write_all(MAGIC_NO_META)?;
         file.write_all(&[1u8])?;
         file.write_all(crc!(bytes).as_slice())?;
         file.write_all(bytes)?;
-        file.flush()?;
+        file.sync_all()?;
+        drop(file);
+
+        fs::rename(&path_tmp, &path)?;
+        sync_dir(base_path)?;
```
sebadob: That's also solved in #357 once the git conflict is solved. Makes sense.

hiqlite-wal/src/writer.rs
```diff
@@ -257,15 +294,36 @@ fn run(
...
+                            flush_and_release(
+                                &mut wal,
+                                &mut buf,
+                                &mut is_dirty,
+                                &mut pending_callbacks,
+                            )?
+                        }
+                    }
+                    // Released by `Action::Sync` on the next tick.
+                    LogSync::IntervalMillis(_) => {}
```
sebadob: The changes for `ImmediateAsync` and `IntervalMillis` completely kill them, and make them worthless.

`ImmediateAsync` is now the exact same performance as `Immediate`, so it would not even make sense anymore to have it, while the bench for `IntervalMillis` gets worse the higher the interval for flushing (which makes sense). Even a 2ms interval is already ~5 times slower than an `Immediate` sync.

I know that `ImmediateAsync` and `IntervalMillis` only did an msync before. That was on purpose. Just for reference: The `Immediate` performance is ~35x lower with a concurrency of `16` than `ImmediateAsync` (with the code before).

In the real world, `msync` should most often be safe enough. The data survives an application crash. Only if the kernel panics (you have way bigger issues if this happens), or if you have a power outage without UPS, you could lose data. But when that happens, your whole node is gone. It will do a crash recovery with the next restart and then sync data from the other cluster members.

---

It makes sense to have stricter syncing and flushing in all other cases like metadata (probably also depending on the set sync level), but not for `ImmediateAsync` and `IntervalMillis`. It removes the possibility from the user to configure it to their needs. If you run a single instance (does not make much sense to run a Raft then) or all nodes on the same physical hardware, then yes of course you probably should switch to `Immediate` and have the strict syncing. 

However, it does not make much sense to run a Raft Cluster in these situations in the first place, as it would not be true HA anyway. But, even if you want to do something like that, you can configure `Immediate` syncing and you are fine.

The documentation for the config setting in the example config file is very clear about this already, even though I admit the same doc is missing for the in-code documentation on the `NodeConfig` struct and needs an update:

```toml
# Setting for Raft Log syncing / flushing.
#
# This value is probably the most important, depending on your needs.
# It has a huge impact on both performance and consistency.
#
# You can set 3 different levels:
# - immediate
# - immediate_async
# - interval_<time_ms> (e.g. interval_200)
#
# If you run a single instance "Cluster", you most probably want
# `immediate` to have the highest degree of consistency. If set
# to `immediate`, the Raft will block until data has been flushed
# to disk. This is especially important for a single instance
# deployment, because there is no way to recover state from other
# nodes or re-sync a maybe corrupted WAL file.
# `immediate` has a very huge negative impact on throughput, and
# it puts a lot of stress on your disk, which is important to
# consider for SSDs.
#
# `immediate_async` puts the same amount of stress on your SSD
# and flushes all buffers to disk immediately after a single
# Raft Log was saved. Unlike `immediate` however, it does not
# wait for completion and directly returns `success` to the
# Raft Engine. You have a bit less consistency guarantees in
# exchange for basically the same throughput as with `interval`
# syncing.
#
# The `interval_<ms>` option will not flush immediately. Flushes
# will be triggered by an external ticker task to flush buffers
# every <ms> ms, and then only if buffers are dirty, meaning if
# any data has been updated since the last sync. This setting
# has the highest throughput and puts the lowest amount of stress
# on your SSD.
# CAUTION: You probably never want to use this setting for a
# single instance, since it may be the case that if you are
# unlucky and your app crashes before buffers are synced, that
# your WAL file might end up being corrupted. Even though very
# unlikely in real world (can be force-produced though), you
# would need to re-sync from a healthy cluster member in such
# a case, if the automactic WAL repair does not succeed.
#
# default: immediate_async
# overwritten by: HQL_LOG_SYNC
log_sync = "immediate_async"
```

sebadob (changes requested):

hiqlite-wal/src/writer.rs
drmingdrmer: Reverted: `ImmediateAsync` and `IntervalMillis` fire the callback immediately again, and the metadata part is gone since #357 landed it. What is left touches only paths where no append waits on the flush:

- `Action::Remove` and `Action::Vote` flush blocking. `is_dirty` now means "the WAL holds bytes that are not known to be on disk": every append sets it, and only a flush that returns clears it. Without that invariant the `if is_dirty` guard in `Action::Remove` was dead under `ImmediateAsync`, because that mode never set the flag — the header was not written before the logs were deleted, which is the hole the comment there guards against. `Action::Vote` cleared the flag right after starting an msync, so the ticker then skipped a WAL that had not reached disk.
- `Action::Sync` flushes blocking too. `flush_async` is `msync(MS_ASYNC)`, which starts no writeback on Linux at all: `mm/msync.c` calls `vfs_fsync_range` only for `MS_SYNC`. The ticker was flushing nothing, so `interval_<ms>` was waiting for normal kernel writeback, 30s by default via `dirty_expire_centisecs`. No append waits on the ticker, so no commit gets slower.
- `flush_blocking` also refreshes the file header before flushing, which the `Immediate` append path skipped.

Together that is one fsync per purge and per vote change, and nothing on the append path.

Why the batching produced your numbers: openraft 0.9 awaits `append()` once per client write, and our `append()` awaits the writer's ack, so the next `Action::Append` cannot have arrived by the time the writer tests `rx.is_empty()`. The batch was always one entry, which is exactly `Immediate`. Batching needs a gather window rather than an emptiness test, and I will bring that separately with numbers.

The `NodeConfig::wal_sync` docs are in as well. One claim in that text is worth your eye, because it cuts against re-syncing from the other members: the loss is not confined to the node that crashed. A leader and one follower hold an entry in the page cache, the leader commits it and answers the client, then power-cycles without it; a third node whose log ends before that entry can win the next election and truncate it from the follower. The acknowledged write is then gone cluster-wide and no member can re-sync it. Reword as you see fit.

If you want `immediate_async` to do what its doc says at no latency cost, `sync_file_range(SYNC_FILE_RANGE_WRITE)` starts writeback without waiting. Happy to add it.

hiqlite-wal/src/writer.rs
```diff
@@ -353,23 +373,19 @@ fn run(
...
-                if is_dirty {
-                    let active = wal.active();
-                    buf.clear();
-                    active.update_header(&mut buf)?;
-                    active.flush_async()?;
-                    is_dirty = false;
-                }
+                // The ticker is the only flush in `IntervalMillis` mode and no append waits on
+                // it, so it blocks: `msync(MS_ASYNC)` starts no writeback on Linux at all.
+                flush_blocking(&mut wal, &mut buf, &mut is_dirty)?;
```
sebadob: While it's true what the comment says, doing a blocking sync here does not improve any guarantees or crash resistance. The ticker runs async, independently, and all logs appended when running in interval sync mode have been reported as being written anyway already. They (might) only exist in the kernel buffers at this point, and the Raft state has moved forward already anyway.

Actually, the `Action::Sync` would not even need to update the header here, but it also does not hurt the performance too much with a ms-internal. If we would crash, the header is being rebuilt after and inspection of all files on disk anyway to guarantee that everything is consistent with what exists on disk.

Doing a blocking sync here degrades the performance even with a 200ms interval below the one of `ImmediateAsync`, which does not even make much sense, when it syncs way less. It performs slightly worse. 

You are correct of course that by definition, `ImmediateAsync` and `InteralMillis(_)` behave wrongly. In the real worl however, when you run on proper hardware, you never lose data. Even when you run 3 nodes on consumer hardware and maybe an instable kernel (I hope not ^^), you won't lose data. If a node crashes, it goes into recovery mode with the next startup, makes sure the header files and metadata match what exists on disk, and then syncs possibly lost data from the other nodes. 

If you lose all 3 nodes at once because of a global power outage or UPS failure (which would be a SPOF and a bad design), then yes, you could lose data. That the kernel panics, and then even on all nodes at once, will not happen.

The Interval sync idea came up mostly because of running clusters on Raspberry Pis back then, which only have insanely slow SD cards. The asnyc flushing brings back performance there, but it kills the SD card in basically no time when you have high traffic. The async flushing is usually so quick that when calculating network latency, it is durable before the data arrives in the state machine writer. And the state machine can never get out of sync as well, because it's rebuilt entirely from the WAL files when it detects a restart after it crashed beforehand.

In the end, it's really up to the user. Stricly correct in terms of the Raft definition would only be the `Immediate` variant, while you can gain something around a 35x perf improvement with the async methods, which have basically the same consistency on a proper server. Interval exists to put less stress on your disks.

While thinking about it, the `Sync` variant could be renamed to `IntervalSync` to make it very clear by the name already that it's only about the interval syncing.

sebadob (commented):
> is_dirty now means "the WAL holds bytes that are not known to be on disk"

Yeah that was exactly the main idea for it. Setting it wrongly was a logic bug.

> Action::Vote cleared the flag right after starting an msync, so the ticker then skipped a WAL that had not reached disk.

I think that could have been the reason for a bug I saw sometimes during rolling updates, which I had no explanation for yet.

> (...) normal kernel writeback, 30s by default via dirty_expire_centisecs. 

I was not aware of the fact that it's such a long ticker. In that case, with blocking flush from `IntervalMills`, it has an even lower chance than `ImmediateAsync` to lose anything because of a power outage, and it would probably be a better default than `ImmediateAsync` with a value like `200ms` or something like that. I remember that Postgres also did something like that, but haven't looked at the internals in a long time.

> flush_blocking also refreshes the file header before flushing, which the Immediate append path skipped.

Makes sense, even though the header is being rebuilt from the data that actually exists on disk, if an unclean restart is detected.

> the loss is not confined to the node that crashed. A leader and one follower hold an entry in the page cache, the leader commits it and answers the client, then power-cycles without it; a third node whose log ends before that entry can win the next election and truncate it from the follower. The acknowledged write is then gone cluster-wide and no member can re-sync it. Reword as you see fit.

That's true, but if the leader dies in the process, the client will not get an `Ok(_)` response anyway, which will come as response from the state machine writer back through the same node. I guess, technically it might be possible if the 3rd node would be so slow that all of this happens before it ever receives the WAL append, but wouldn't it lose a possible election in that case because any quorum node would have a higher committed log id?

> If you want immediate_async to do what its doc says at no latency cost, sync_file_range(SYNC_FILE_RANGE_WRITE) starts writeback without waiting. Happy to add it.

This means, it would be basically flush async, but with an immediate trigger instead of waiting for the kernel ticker? That would be nice.

In did some bench comparison with the current branch, and Interval syncing has a tiny bit less throughput than ImmediateAsync, which makes sense with all that knowledge and the blocking sync. It is behind ImmediateAsync by ~3.6% when inserting 1_000_000 rows. If we can trigger an immediate async, that would bring that 3.6% back. Realistically, this is already at a rate of DB INSERTS at which you would run a dedicated Postgres server only for the DB, but getting the perf back would be nice, because we don't gain any consistency by blocking the sync here.

drmingdrmer:
> if the leader dies in the process, the client will not get an `Ok(_)` response anyway

The `Ok` and the power loss are separated by the writeback window. The leader answers as soon as a quorum acks; the crash can come at any point in the following ~30s while the entry sits in page cache. The client holds a confirmed write that whole time.

> wouldn't it lose a possible election in that case because any quorum node would have a higher committed log id?

Vote checks compare last log id — the committed id is not consulted. And the electing quorum is the restarted leader plus the third node, and the restarted leader's log has regressed: L and F1 ack E@10 into page cache, F2 ends at 9; L commits, answers the client, then power-cycles before writeback and before its append reaches F2. L restarts ending at 9, so it grants F2's vote request (9 ≥ 9). {L, F2} is a quorum; F1's refusal is outvoted, and the new term's leader log ends at 9, so F1 truncates E on first contact. The acknowledged entry is gone from every node.

You're right that this variant needs the crash to land before F2 receives E — milliseconds, or longer only when F2 lags for mundane reasons. The variant the doc paragraph is really about needs no election at all: L and F1 lose power together (same PSU, rack, or hypervisor — the "all nodes on the same physical hardware" case from your config doc), and E is on no disk afterwards. That one has the full writeback window.

> I think that could have been the reason for a bug I saw sometimes during rolling updates

One caveat before you close that hunt: dirty mmap pages live in the kernel page cache, which survives SIGKILL and process restarts, and a clean reboot flushes on unmount. The skipped flush loses data only on power loss, kernel panic, or hard VM reset. Unless those rolling updates hard-reset machines, that bug is still unexplained.

> This means, it would be basically flush async, but with an immediate trigger instead of waiting for the kernel ticker?

Exactly that. Two properties to settle before I wire it in:

- It is not a durability barrier: it neither flushes the device write cache nor commits metadata. WAL files are created sparse (`create_new` + `set_len`), so the extent allocation for a freshly written range becomes durable only with the next filesystem journal commit — after a power loss such a range can read back as zeros even though the device received the pages. So it never clears `is_dirty`, and `Remove`/`Vote` keep their blocking flush.
- For the ticker: replacing the blocking flush with it turns "at most `<ms>` of unsynced data" into "typically `<ms>`, no hard bound". That bound is the only thing the blocking tick buys — your call whether it is worth the 3.6%.

I'll add it to `immediate_async`: start writeback after each append, `#[cfg(target_os = "linux")]` with `flush_async` as the fallback — macOS and Windows actually start I/O on `flush_async`, the no-op is Linux-specific. That also answers the default question: with per-append writeback start, `immediate_async`'s typical exposure is a few milliseconds, smaller than any tick, so it stays the stronger default.

sebadob:
> I'll add it to immediate_async: start writeback after each append, #[cfg(target_os = "linux")] with flush_async as the fallback — macOS and Windows actually start I/O on flush_async, the no-op is Linux-specific. That also answers the default question: with per-append writeback start, immediate_async's typical exposure is a few milliseconds, smaller than any tick, so it stays the stronger default.

That's really helpful, and also answers why I saw the exact same performance for `ImmediateAsync` and `IntervalMillis`. With the current state of this PR, `ImmediateAsync` has reduced throughput by ~24%, but it has way higher possibility of not loosing data on a power outage.

The only thing I would add is a comment why the `unsafe` sync is safe, but I can do that with the next release, no problem.

Thank you!

sebadob (approved):

PR #362, opened 2026-08-17
feat(cache): atomic get_remove() and replace() operations

reneleonhardt:
## Summary

Adds `get_remove` and `replace` as atomic cache operations, so a value can be claimed or swapped in a single raft entry instead of a `get` + `delete` / `get` + `put` pair.

- `get_remove` returns a value and removes it in one step. Concurrent callers race on the raft log, not on the read-modify-write window: exactly one observes the value, all others get `None`. This is what single-use claims need (one-time tokens, PoW challenges, logout-token replays) — the two-step variant let two callers both observe the value.
- `replace` stores a new value and returns the previous one in one step, for atomic renewals where the caller must know the old value (e.g. refresh-token families).

Both are raft log entries applied on all nodes like any other cache write, and the actual map mutation happens on the single KV handler thread, so there is no get+delete window to race.

## Changes

- `Client::get_remove` / `get_remove_bytes`: return and remove in one raft entry.
- `Client::replace` / `replace_bytes`: store a new value and return the old one; optional TTL with the same contract as `put` (`None` = no expiry, a previously registered expiry for the key is dropped).
- TTL handling mirrors `put`/`delete` exactly: a `replace` with a TTL re-registers the expiry, a re-put without one clears it, and a removed entry leaves no stale expiry behind.
- KV handler unit tests plus cluster and remote-client probes pin the semantics: single-claim under concurrent `get_remove`/`replace` races, cross-node visibility, old-value return, both TTL directions, and replay of `get_remove` entries after a snapshot purge restart.

## Review follow-up

- `replace()` no longer applies the cache rate limit twice; it relies on the one inside `replace_bytes()`.
- `CacheRequest` and `QueryWrite` variants are no longer feature-gated, so the serialized variant order is identical for every feature set and raft logs stay readable when features change between deployments. Disabled-feature match arms use `unreachable!()`, gated variants carry `#[allow(dead_code)]`, and the order is pinned by unit tests so a future reorder fails CI instead of corrupting logs.

## Verification

- `cargo test -p hiqlite --lib --features cache,counters,dlock,listen_notify,macros,toml` — 42 passed.
- `cargo test -p hiqlite --test cluster --features cache,counters,dlock,listen_notify,macros,toml` — all tests successful (cluster, self-heal, remote-only, including the new probes).
- `cargo clippy -- -D warnings` and the no-default feature matrix — clean.
- Regression-proven: the new probes fail against a simulated non-atomic get-without-remove implementation and pass on the real one.

## Compatibility notes

The `get_remove` / `replace` API itself is additive. The enum de-gating is a deliberate breaking change requested in review: the serialized index of `ClearAll` (and the `counters`/`dlock`/`listen_notify_local` variants) in `CacheRequest` and of `RTT` in `QueryWrite` can shift between builds with different feature sets, so an upgrade to this version requires clearing the cache and starting from fresh WAL files (new minor version).

### DIFF
--- a/hiqlite/src/client/cache.rs
+++ b/hiqlite/src/client/cache.rs
@@ -306,8 +306,7 @@ impl Client {
         K: Into<Cow<'static, str>>,
         V: Serialize + for<'a> Deserialize<'a>,
     {
-        self.rate_limit_cache().await?;
-
+        // `replace_bytes` below applies the cache rate limit itself
         match self
             .replace_bytes(cache, key, serialize_network(value), ttl)
             .await
--- a/hiqlite/src/client/cache.rs
+++ b/hiqlite/src/client/cache.rs
@@ -245,6 +245,119 @@ impl Client {
         Ok(())
     }
 
+    /// GET and REMOVE a value from the cache in one atomic step.
+    ///
+    /// Of several concurrent `get_remove` calls for the same key, exactly one
+    /// returns the value; all others get `None`.
+    pub async fn get_remove<C, K, V>(&self, cache: C, key: K) -> Result<Option<V>, Error>
+    where
+        C: CacheVariants,
+        K: Into<Cow<'static, str>>,
+        V: for<'a> Deserialize<'a>,
+    {
+        match self.get_remove_bytes(cache, key).await {
+            Ok(value) => {
+                if let Some(v) = value {
+                    Ok(Some(deserialize(&v)?))
+                } else {
+                    Ok(None)
+                }
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    /// GET and REMOVE a raw bytes value from the cache in one atomic step.
+    ///
+    /// Works in the same way as `.get_remove()` without any value mapping.
+    pub async fn get_remove_bytes<C, K>(&self, cache: C, key: K) -> Result<Option<Vec<u8>>, Error>
+    where
+        C: CacheVariants,
+        K: Into<Cow<'static, str>>,
+    {
+        self.rate_limit_cache().await?;
+
+        let res = self
+            .cache_req_retry(
+                CacheRequest::GetRemove {
+                    cache_idx: cache.hiqlite_cache_index(),
+                    key: key.into(),
+                },
+                false,
+            )
+            .await?;
+        match res {
+            CacheResponse::Value(opt) => Ok(opt),
+            _ => unreachable!(),
+        }
+    }
+
+    /// REPLACE a value in the cache and return the previous value (if the key existed).
+    /// The optional `ttl` is the lifetime of the value in seconds from *now* on.
+    pub async fn replace<C, K, V>(
+        &self,
+        cache: C,
+        key: K,
+        value: &V,
+        ttl: Option<i64>,
+    ) -> Result<Option<V>, Error>
+    where
+        C: CacheVariants,
+        K: Into<Cow<'static, str>>,
+        V: Serialize + for<'a> Deserialize<'a>,
+    {
+        self.rate_limit_cache().await?;
+
+        match self
+            .replace_bytes(cache, key, serialize_network(value), ttl)
+            .await
+        {
+            Ok(value) => {
+                if let Some(v) = value {
+                    Ok(Some(deserialize(&v)?))
+                } else {
+                    Ok(None)
+                }
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    /// REPLACE a raw bytes value in the cache and return the previous value.
+    pub async fn replace_bytes<C, K>(
+        &self,
+        cache: C,
+        key: K,
+        value: Vec<u8>,
+        ttl: Option<i64>,
+    ) -> Result<Option<Vec<u8>>, Error>
+    where
+        C: CacheVariants,
+        K: Into<Cow<'static, str>>,
+    {
+        self.rate_limit_cache().await?;
+
+        let res = self
+            .cache_req_retry(
+                CacheRequest::Replace {
+                    cache_idx: cache.hiqlite_cache_index(),
+                    key: key.into(),
+                    value,
+                    expires: ttl.map(|seconds| {
+                        Utc::now()
+                            .timestamp_micros()
+                            .saturating_add(seconds.saturating_mul(1_000_000))
+                    }),
+                },
+                false,
+            )
+            .await?;
+        match res {
+            CacheResponse::Value(opt) => Ok(opt),
+            _ => unreachable!(),
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


hiqlite/src/client/cache.rs
```diff
@@ -245,6 +245,119 @@ impl Client {
...
+        key: K,
+        value: &V,
+        ttl: Option<i64>,
+    ) -> Result<Option<V>, Error>
+    where
+        C: CacheVariants,
+        K: Into<Cow<'static, str>>,
+        V: Serialize + for<'a> Deserialize<'a>,
+    {
+        self.rate_limit_cache().await?;
```
sebadob: Duplicate rate-limiting here. It calls `replace_bytes()` which already rate-limits.

hiqlite/src/store/state_machine/memory/state_machine.rs
```diff
@@ -67,6 +67,16 @@ pub enum CacheRequest {
...
+    GetRemove {
+        cache_idx: usize,
+        key: Cow<'static, str>,
+    },
+    Replace {
+        cache_idx: usize,
+        key: Cow<'static, str>,
+        value: Vec<u8>,
+        expires: Option<i64>,
+    },
```
sebadob, line 83: This brings us an issue with backwards compat. 

However, we could do it like that now, while at the same time improving something else I had on my TODO since a while:

All `enum`s that are sent over the Raft, and therefore get serialized into the WAL file and Raft Logs, have an issue when enabling / disabling features with an existing dataset. The problem is the deserilization in combination with the variants that only exist depending on enabled features. During serialization, the variants are assigned to an integer under the hood. The problem is, when they are deserialized with a different feature set and therefore different ordering, it will lead to wrong variants with wrong values than intended.

My idea back then was to enable only the variants depending on features, if they are actually needed, but this is actually a bad idea because of the above problem. To not break backwards compat, the new variants would need to be added at the end. Even though this would work now, it's not the best solution right now. When we are adding new variants anyway, we should tackle the mentioned issue for DX improvments.

The variants and order should not change depending on activated features. This is true for the cache and db state machines, but only for the ones getting into the Raft logs. When we remove the flags like `#[cfg(feature = "counters")]` and make sure they always contain the same variants in the same order, everything will keep working even when you change the features for an existing app. It would require a minor version release, and it's a big breaking change, but it's also a big improvement for the future. When you start the new version, you must make sure that you clear all caches and you don't build up on old WAL files.

The DB side is using `QueryWrite` here. This enables a single variant depending on the `backup` feature, and afterwards there is only `RTT` which could get out of order when an existing DB did not have the `backup` feature enabled. `RTT` however is only used during backup restore, so nothing to worry about. We can remove the feature gate without breaking an existing DB.

The Cache side is using `CacheRequest` which has many feature gates. However, it's only a cache, and clearing it for an upgrade should not be an issue for existing apps.

---

So, to tackle this situation, I would keep the ordering like you did it, because it makes sense logically. This is a breaking change though and needs a cache clear. To avoid future clears, we should also remove the feature gates on the other enum variants at the same time. This might trigger some clippy warnings depending on activates features, but we can `allow(unused)` these situations specifically. We can even make use of `unreachable!()` in these places. 

This is a small overhead, but we get actually less cluttered enum definitions and the possibility to change features without breaking an existing deployment in the future. We only need to keep the correct padding for serialized enums independent from enabled features. Feature gates in all other places should stay like they are of course for reduced code size.

sebadob (changes requested):

reneleonhardt:
GitHub is so bad. It shows you requested 1 change. But which one? 😄

Edit: Now inline comments are shown, 9 minutes after creation. GitHub is soo down since 24h 😞

sebadob:
> GitHub is so bad. It shows you requested 1 change. But which one? 😄
>
> Edit: Now inline comments are shown, 9 minutes after creation. GitHub is soo down since 24h 😞

Yeah it's a nightmare. So you see both of them now?

Edit:

I almost lost the big review I did yesterday on Rauthy. Right after submitting, Github went down for hours on my side.

reneleonhardt:
Yeah, I got them. For now.

Tell me about it, all my agents were blocked in all network layers with different error response codes, it still is a nightmare, and without even an official response or apology for that global outage 😞

sebadob:
> Tell me about it, all my agents were blocked in all network layers with different error response codes, it still is a nightmare, and without even an official response or apology for that global outage 😞

If it was not for exposure for open source projects, I would be over on codeberg already. I do use it for some private repos. It's just so much more reliable. I am also pretty sure the Github status page shows fake data. I personally have issues more often than this page showing any.

reneleonhardt:
I couldn't agree more, I only look at the status when I detect outage myself, always green in GitHub land 💯

reneleonhardt:
### Compliance

- Comment 1 (duplicate rate limiting): fixed — `replace()` now rate-limits exactly once via `replace_bytes()`.
- Comment 2 (feature-gated raft enums): done — all gates removed from `CacheRequest` and `QueryWrite`, ordering kept as you suggested, disabled-feature arms use `unreachable!()`, gated variants carry `#[allow(dead_code)]`. Feature gates everywhere else stay.

### Where I didn't fully comply

- None — both comments are fully addressed.

### Other changes the review forced (not in your suggestions)

- The de-gating makes this a breaking change: `ClearAll` and later `CacheRequest` variants shift index for builds without `counters`/`dlock`/`listen_notify_local`, `RTT` in `QueryWrite` for builds without `backup`. Upgrading requires a cache clear and fresh WAL files — noted in the PR description.
- Two unit tests pin the serialized variant order, so a future reorder fails CI instead of corrupting raft logs.
- Both enums carry a contract comment now: new variants go at the end.

sebadob (approved):
Thanks!

With this on `main`, you can reference it in the Rauthy PRs. The crates io patch is already inside the `Cargo.toml`, out-commented. I will take care of releasing a stable `hiqlite` version and switching the reference before the next Rauthy release.

Edit: 

In the Rauthy PR, making sure in `src/data/src/temp_migrations.rs` that the cache WAL is cleaned up is an important step for the whole next minor release. Since the original PR is so big already, adding this in a 2nd one makes the most sense. It can then also bump the minor version with a nightly and remove the current temp migrations. They are only ever valid for one minor version, so users don't have to do anything manually when upgrading.

reneleonhardt:
Thanks! The rauthy PR will pick this up: the crates-io patch is activated, the three dlock claim sites (PoW, authorization code, device code) switch to `get_remove()`, and the `dlock` feature drops out of the dependency.

Agreed on the temp_migrations cache-WAL cleanup — that goes into a follow-up rauthy PR together with the nightly minor bump and removing the temp migrations.

PR #364, opened 2026-08-31
feat(sqlite): add an externally committed state-machine engine

eljeffeg:
## Motivation

Some embedded users already own a trusted consensus log and need Hiqlite's SQLite single-writer, read pool, snapshot,
and restore machinery without creating a second Raft group. Driving the existing Raft `TypeConfig` from another log
would couple two authorities and expose the raw-SQL `QueryWrite` wire contract.

This PR adds an opt-in, engine-only seam. It creates no Hiqlite Raft node, network service, membership metadata, or
raw-SQL replication protocol. The caller remains responsible for consensus, canonical command encoding, deterministic
typed operations, replay, and its outer snapshot manifest.

The feature also owns its small SQLite connection/restore helpers directly and does not activate the regular
Hiqlite cluster dependency boundary. This keeps OpenRaft, hiqlite-wal, networking, TLS, S3, and dashboard code out of
an external-only consumer and prevents the caller's OpenRaft features from unifying into dormant Hiqlite Raft code.

## Public contract

The new `external-state-machine` feature is excluded from `default` and `full`. Its key types are:

- `CommitSequence(u64)`: the explicit dense monotonic order. Caller-native coordinates are opaque `Eq` identities and
  are never used through a generic `Ord` contract.
- `ExternalCommit<C>`: sequence, exact native coordinate, SHA-256 canonical command digest, state schema, and receipt
  schema.
- `DeterministicSqliteOperation`: a caller-defined typed operation and typed serializable output. It receives the exact
  re-exported `rusqlite::Transaction`; an authorizer denies writes to reserved metadata, transaction control, PRAGMAs,
  ATTACH/DETACH, and temporary objects. Protocol owners can override strict `encode_receipt`/`decode_receipt` hooks for
  a canonical durable receipt format; strict legacy bincode remains the compatibility default.
- `apply_committed` and `advance_committed`: every global entry must use one. Advance-only entries persist the common
  applied frontier for entries that do not mutate this SQLite lane.
- `ApplyOutcome::{Applied, Recovered}`: mutation, frontier, and a bounded typed receipt commit in one SQLite transaction.
  An exact retained retry returns the stored response without re-executing. Reusing the same sequence with different
  coordinate/digest/schema/kind/codec conflicts; a pruned response returns `ReceiptUnavailable`.
- `ExternalSnapshotEvidence<C>`: exact page-image format/version, application/schema identity, receipt window, caller
  checkpoint, page size, length, and SHA-256. It intentionally contains no membership. Its Serde shape is descriptive,
  not a stable wire-format promise: callers persist a versioned outer representation they own, then reconstruct typed
  evidence with `ExternalApplied::new` and `ExternalSnapshotEvidence::from_manifest` for validation or install.
- `EXTERNAL_SQLITE_SNAPSHOT_FORMAT`: the public page-image format identifier for outer manifests, so callers do not
  duplicate an engine-private literal.

The API does not claim atomicity between SQLite and the caller's log or another storage engine. Recovery compares the
stored frontier with the durable caller log and replays.

## Durability and recovery

`ExternalDurability::Full` is the safe default. A successful FULL apply/advance acknowledges after SQLite commits the
WAL transaction containing the application mutation, frontier, and receipt. `Normal` is explicit and may lose an
acknowledged tail on power loss, so the caller must replay. `ReplayableOff` is also explicit; any unclean OFF marker
requires `rebuild_projection`, even if the next open requests FULL/NORMAL.

External mode takes an advisory owner lock and keeps a separate durable mode/restore marker. It never invokes
`auto-heal` and `open` never deletes or replaces the database. Only an explicit `rebuild_projection` is allowed, and
only after a rebuild-required marker. Clean shutdown upgrades every mode to FULL, checkpoints/truncates WAL, closes
SQLite, syncs the database and parent directory on Unix, removes and directory-syncs the marker, then releases the
owner lock.

External read-pool connections use `SQLITE_OPEN_READ_ONLY`, not only `query_only`. The regular clustered connection,
read-pool, and restore paths are unchanged by this feature.

## Snapshots

External snapshots use SQLite Online Backup behind the single-writer queue. This gives one exact applied frontier and
blocks later writes for the copy while pooled reads continue. A sibling temporary file is published without replacing
an existing final path only after the copy completes, then synced before evidence is returned. Restore validates
format/application/frontier/receipt window, length, SHA-256, page size, and `quick_check` before touching the live
database, then rechecks staleness inside the writer queue. A failure after live restore begins poisons the writer and
leaves a rebuild-required marker. `ExternalSqlite::validate_snapshot` exposes that complete file/embedded-metadata
check without opening or mutating a live engine, so a caller can verify its decoded outer-manifest evidence before
activation. The caller's outer manifest should separately bound and authenticate its own encoding; the engine validates
the reconstructed typed evidence and SQLite artifact rather than owning that outer wire format.

I chose a page-image backup rather than reusing the existing `VACUUM INTO` helper because SQLite documents that VACUUM
may change implicit ROWIDs. A local 200k-row/256-byte-payload benchmark measured median queue pauses of 40 ms for
VACUUM and 61 ms for Online Backup; the compact VACUUM output was 27.4 MB versus the 54.8 MB page image. The generic
state-machine identity guarantee outweighs that local speed/space win. The caller owns completed-path retention and
durably recording the returned evidence in its outer manifest; that outer step is not atomic with hiqlite.

## Validation

Validated branch head: `7488d3c00690806b929d5b6f216f90fdf6a186eb` on upstream base `c3ff253`. The final
PR diff is five focused files; regular clustered connection, read-pool, and restore sources are unchanged from base.

- Rust 1.96: `cargo test -p hiqlite --lib --no-default-features --features external-state-machine` — 9 test functions passed
- Rust 1.96: `cargo test -p hiqlite --lib` — 31 passed
- Rust 1.96: `cargo test -p hiqlite --lib --features external-state-machine` — 40 passed
- Rust 1.96 strict external clippy passes with the repository's pre-existing `result_large_err` lint allowed
- Nightly `cargo doc --no-deps --no-default-features --features external-state-machine` with `RUSTDOCFLAGS=-Dwarnings`
- `git diff --check`

The nine new regression tests cover ordering/advances, retry/conflict/pruning, transaction rollback, receipt limits,
caller-owned codecs, reserved-metadata isolation/read-only pooling, explicit adoption, snapshot evidence and restore
validation/read-pool replacement, implicit ROWIDs, durability modes, owner locking, dirty/poison markers, explicit
rebuild, crash reopen, and clean shutdown.

## Compatibility and scope

- Coordinates use Hiqlite's strict legacy-bincode configuration. Receipts use the operation's selected codec (strict
  legacy bincode by default). Coordinate/receipt formats and `RECEIPT_CODEC` must remain backward-decodable for retained
  receipts and restorable snapshots; `receipt_schema` is evidence rather than decoder dispatch in format v1.
- Hiqlite's existing Raft snapshot/backup/S3 behavior remains unchanged.
- Dependency impact is opt-in and limited to the external SQLite engine: bincode, deadpool, flume, fs4, rusqlite,
  serde, SHA-256, error/thread/runtime support, tracing, and UUID. It does not select Hiqlite's OpenRaft,
  hiqlite-wal, network, TLS, backup/S3, dashboard, or server graph.
- License remains Apache-2.0.

### DIFF
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -23,7 +23,7 @@ default = ["auto-heal", "backup", "sqlite", "toml"]
 
 auto-heal = ["hiqlite-wal/auto-heal"]
 backup = ["dep:cron", "s3", "sqlite"]
-cache = ["cluster", "openraft/loosen-follower-log-revert"]
+cache = ["__cluster", "openraft/loosen-follower-log-revert"]
 counters = ["cache"]
 dashboard = [
     "dep:argon2",
@@ -87,7 +87,7 @@ server = [
 ]
 shutdown-handle = ["dep:ctrlc"]
 sqlite = [
-    "cluster",
+    "__cluster",
     "dep:deadpool",
     "dep:rusqlite",
     "dep:serde_rusqlite",
@@ -98,7 +98,7 @@ webpki-roots = [
 
 # Internal dependency boundary for the regular Hiqlite Raft/network runtime.
 # The external state-machine feature deliberately does not enable it.
-cluster = [
+__cluster = [
     "dep:axum",
     "dep:axum-server",
     "dep:bytes",
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -16,7 +16,7 @@ doctest = false
 
 [package.metadata.docs.rs]
 all-features = false
-features = ["full"]
+features = ["full", "external-state-machine"]
 
 [features]
 default = ["auto-heal", "backup", "sqlite", "toml"]
@@ -37,6 +37,10 @@ dashboard = [
 dlock = ["cache"]
 cast_ints = []
 cast_ints_unchecked = []
+# Opt-in state-machine engine for applications that already own consensus.
+# This reuses the SQLite connection, read-pool, restore, and single-writer
+# machinery without starting a Hiqlite Raft group.
+external-state-machine = ["sqlite", "dep:fs4", "rusqlite/hooks"]
 full = [
     "auto-heal",
     "backup",
@@ -111,6 +115,7 @@ dotenvy.workspace = true
 eventsource-client = { workspace = true, optional = true }
 fastwebsockets.workspace = true
 flume.workspace = true
+fs4 = { workspace = true, optional = true }
 futures-util = { workspace = true, optional = true }
 getrandom.workspace = true
 hex.workspace = true
--- a/hiqlite/Cargo.toml
+++ b/hiqlite/Cargo.toml
@@ -23,7 +23,7 @@ default = ["auto-heal", "backup", "sqlite", "toml"]
 
 auto-heal = ["hiqlite-wal/auto-heal"]
 backup = ["dep:cron", "s3", "sqlite"]
-cache = ["openraft/loosen-follower-log-revert"]
+cache = ["cluster", "openraft/loosen-follower-log-revert"]
 counters = ["cache"]
 dashboard = [
     "dep:argon2",
@@ -40,7 +40,14 @@ cast_ints_unchecked = []
 # Opt-in state-machine engine for applications that already own consensus.
 # This reuses the SQLite connection, read-pool, restore, and single-writer
 # machinery without starting a Hiqlite Raft group.
-external-state-machine = ["sqlite", "dep:fs4", "rusqlite/hooks"]
+external-state-machine = [
+    "dep:deadpool",
+    "dep:fs4",
+    "dep:rusqlite",
+    "rusqlite/backup",
+    "rusqlite/functions",
+    "rusqlite/hooks",
+]
 full = [
     "auto-heal",
     "backup",
@@ -80,12 +87,49 @@ server = [
 ]
 shutdown-handle = ["dep:ctrlc"]
 sqlite = [
+    "cluster",
     "dep:deadpool",
     "dep:rusqlite",
     "dep:serde_rusqlite",
 ]
 webpki-roots = [
     "dep:webpki-root-certs",
+]
+
+# Internal dependency boundary for the regular Hiqlite Raft/network runtime.
+# The external state-machine feature deliberately does not enable it.
+cluster = [
+    "dep:axum",
+    "dep:axum-server",
+    "dep:bytes",
+    "dep:chrono",
+    "dep:crossbeam",
+    "dep:cryptr",
+    "dep:dotenvy",
+    "dep:fastwebsockets",
+    "dep:getrandom",
+    "dep:hex",
+    "dep:hiqlite-wal",
+    "dep:hostname",
+    "dep:http-body-util",
+    "dep:hyper",
+    "dep:hyper-util",
+    "dep:lazy_static",
+    "dep:lz4-sys",
+    "dep:openraft",
+    "dep:rcgen",
+    "dep:reqwest",
+    "dep:rust_decimal",
+    "dep:rust-embed",
+    "dep:rustls",
+    "dep:rustls-webpki",
+    "dep:serde_json",
+    "dep:serde_rusqlite",
+    "dep:time",
+    "dep:tokio-rustls",
+    "dep:tower-layer",
+    "dep:tower-service",
+    "dep:url",
 ]
 
 # Internal use only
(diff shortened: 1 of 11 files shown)


hiqlite/Cargo.toml
```diff
@@ -84,77 +96,114 @@ webpki-roots = [
     "dep:webpki-root-certs",
 ]
 
+# Internal dependency boundary for the regular Hiqlite Raft/network runtime.
+# The external state-machine feature deliberately does not enable it.
+cluster = [
```
sebadob: Should be prefixed with `__`  to make it more obvious it's an internal feature.

hiqlite/src/external_state_machine.rs
```diff
@@ -0,0 +1,3363 @@
...
+//!
+//! Restore stages and validates the full artifact, then rechecks staleness
+//! inside the writer queue. Any failure after live restore begins poisons the
+//! engine and leaves a rebuild-required marker. Backup/S3 orchestration from
+//! Hiqlite's Raft API is intentionally outside this feature.
+
+use deadpool::unmanaged::PoolError;
+use fs4::FileExt;
+/// The exact `rusqlite` version used by this external engine.
+pub use rusqlite;
```
sebadob: Re-exposing `rusqlite` makes sense. I had a conflict in another application already. It would be a good idea to move it to `lib.rs` directly.

sebadob (changes requested):
Pretty good already. Just a few things:

- Please update `just` recipes like `check` and `clippy` to also check the new feature properly. The `test` and `test-no-s3` probably need the new feature as well to not skip them. `just` is used for all the testing and release pipelines.
- If there is not at least a minimal walkthrough example how to set this up and use it properly, probably no one (apart from you) will use it. It's fine if the example just mocks the state engine. Just showing how to get started with it would be good.

Apart from that, there is a little bit of duplicated code, but that's fine for now. I plan to create a fully custom read conn pool that mostly gets rid of internal locking, but I can move it out when necessary.

Serialization will also be updated in the near future, but hopefully we can use `wincode` directly as a drop-in replacement, but I have not looked into that yet.

hiqlite/Cargo.toml
eljeffeg: Done in 7929be9 — renamed to `__cluster` (and updated the two internal references in `cache` and `sqlite`).

hiqlite/src/external_state_machine.rs
eljeffeg: Done in 7929be9 — moved to `lib.rs` as `pub use rusqlite;`, gated on `any(feature = "sqlite", feature = "external-state-machine")` so it is available for the regular build too. The new example uses `hiqlite::rusqlite` to demonstrate it.

eljeffeg:
Thanks for the review! Addressed everything in 7929be9:

**`just` recipes**
- `check`: added `cargo minimal-versions check -p hiqlite --no-default-features --features external-state-machine`
- `clippy`: added four combinations — `external-state-machine` standalone (proves it builds without `__cluster`), `sqlite,external-state-machine`, `full,external-state-machine`, and `default + external-state-machine`
- `test` / `test-no-s3`: `external-state-machine` is now part of the feature list, so the 9 new unit tests run in the CI/release pipeline (`just test-no-s3` ran 54 lib tests locally)

**Example**
- New `examples/external-state-machine` with a mocked in-memory consensus log. It walks through defining a `DeterministicSqliteOperation`, mapping log entries to `ExternalCommit`s with a dense `CommitSequence`, replaying the log (same path for fresh start and recovery), `advance_committed` for non-SQLite entries, exact vs. conflicting retries, building/validating/installing a snapshot on a second node, and clean shutdown + reopen. It builds with `default-features = false` so it only compiles the engine. Linked from the README feature section; `just clippy-examples` picks it up automatically.

**Inline comments**
- `cluster` → `__cluster`
- `pub use rusqlite;` moved to `lib.rs`, gated on `sqlite` or `external-state-machine`

Verified on 1.96: all `just clippy` combinations (plus the four new ones) pass with `-D warnings`, `just test-no-s3` feature set passes, nightly `cargo doc` with `RUSTDOCFLAGS=-Dwarnings` for both `external-state-machine` alone and `full,external-state-machine`, and the example runs end to end.

Noted on the read-pool duplication and `wincode` — happy to rebase onto whichever lands first.

sebadob (approved):
Thanks

PR #368, opened 2026-09-21
Robustness and Security hardening

sebadob:
This PR brings lots of small tweaks and changes for robustness, general code stability, a tiny bit of security. Nothing major, but all combined, they add up.


bartekus:
Hi, I have been running a line-by-line specification pass over hiqlite on a fork, and read through this PR in that context. Most of it I will put in a separate comment as a list, but one thing I think is worth its own thread because it concerns unreleased destructive code and is easier to change now than after it ships.

## The v0.15 cache migration safeguard fails open

`v0_15_auto_cache_migration.rs` deletes `logs_cache/` and `state_machine_cache/` after checking that no peer is still running an old version. The check is a good idea and the changelog is admirably loud about the risk. The failure mode of the check itself is the part I would look at again: **an unreachable peer is treated as a safely shut down peer.**

Concretely, at head `dda4f53`:

- `:139` on the `/ping` request, any connection error hits `continue` and the node is skipped entirely.
- `:174` the `/version` check loops twice, and `:161-163` logs the error and sleeps. If both attempts fail, the loop simply ends, the function proceeds to the next node, and no error is returned.

The comment at `:161` says the reasoning is that a node which went down between `/ping` and `/version` is fine for us, which is true for that case. The problem is that a connection error is indistinguishable from several other causes: a TLS handshake failure, a proxy in front of the API port, a transient partition, or a node that is up and serving but momentarily unreachable. In all of those, every peer is silently classified as safe and the destructive cleanup proceeds, which is exactly the scenario the changelog says must not happen.

This compounds with the next point.

## The same check mixes the two TLS configurations

`:130` takes the scheme from `config.tls_api`, while `:83-87` derives `tls_no_verify` from `config.tls_raft` and builds the client with it:

```rust
let tls_no_verify = config
    .tls_raft                      // :83
    .as_ref()
    .map(|c| c.danger_tls_no_verify())
    .unwrap_or(false);
let client = build_http_client(tls_no_verify);   // :87
...
let scheme = if config.tls_api.is_some() { "https" } else { "http" };   // :130
```

Every request here goes to `node.addr_api`, so both should come from `tls_api`. In a cluster where both endpoints are configured the same way this is invisible. In a cluster with `tls_api` set to specific certificates and `tls_raft` not, or with different `danger_tls_no_verify` values, every probe fails verification, every peer therefore looks unreachable, and by the point above the cleanup proceeds anyway.

(This same `tls_raft`-for-`tls_api` pairing exists in `start.rs:38-43` feeding `become_cluster_member`, and in `client/mgmt.rs:186-190`, while `store/mod.rs:98-109` uses `tls_api` for the same endpoint. I will cover that in the other comment; I mention it here only because the new code inherits the pattern.)

## `in-memory-snapshots` builds can never start without the env var

`:28-30` sets `is_pure_in_mem` from the feature, and `:41` guards the whole "already migrated / fresh startup" block behind `else if !is_pure_in_mem`. With the feature on, neither early return at `:44` or `:60` is reachable, so control always falls through to the gate at `:70` and returns "Read the changelog!".

`in-memory-snapshots = ["cache"]` (`hiqlite/Cargo.toml:67`), so `check_migrate` always runs for those builds. As far as I can follow it, such a node refuses to start on **every** startup for the whole 0.15 line unless `HQL_CACHE_WAL_AUTO_MIGRATE=true` is permanently set, and there is nothing on disk to clean up in that configuration anyway.

## One smaller thing in the same file

`:51` and `:58` probe whether the cache directories are empty with `if let Ok(mut dir) = fs::read_dir(...)`. A `read_dir` failure for a reason other than "missing", a permissions problem for instance, leaves `exit_early` true, so the function concludes "No existing cache data found" and returns `Ok`. That is the same fail-open shape: an error about the data is read as an absence of data.

## What I would suggest

I do not want to prescribe a design for your own migration, but the shape that would close this:

1. Distinguish "connection refused" from TLS, handshake, timeout and proxy errors, and treat only the first as evidence that a node is down.
2. Treat "I could not determine this peer's state" as a **hard stop** requiring an explicit acknowledgement, rather than inferring shutdown from silence. Something like `HQL_CACHE_WAL_AUTO_MIGRATE=true` meaning "clean up if you are sure" and a separate opt-in for "proceed even though N peers were unreachable".
3. Use `tls_api` for both the scheme and the verification flag, since the target is `addr_api`.
4. Distinguish `NotFound` from other `read_dir` errors at `:51` and `:58`.

Happy to open a PR for any of these if it would help, or to leave it with you. Thanks for the work on this PR generally: the panic removal through `init.rs` and the `node_id` lookup fix in `start.rs` both close things I had independently written up on the fork.

bartekus:
Second and last comment, the smaller observations from the same read. All line references are against head `dda4f53`. None of these is urgent; they are the kind of thing that is cheap now and annoying later.

### 1. The changelog describes an implementation this PR no longer contains

`CHANGELOG.md:110-118`, "Cache Collision Safety", describes advancing a conflicting TTL in 1-microsecond steps until it finds a free slot, and warns that inserting a million entries with the same absolute TTL would overflow.

A later commit in this PR replaced that approach with a timestamp-to-key-list index. `kv_handler.rs:135` now reads:

```rust
// timestamp, so no collision bumping is needed anymore.
```

and `collision_bump_keeps_both_keys` is gone. The new design looks better to me, but the changelog is user-facing and currently documents a limitation that no longer exists.

### 2. `todo!()` in library startup

`v0_15_auto_cache_migration.rs:24`:

```rust
#[cfg(debug_assertions)]
if !APP_VERSION.starts_with("0.15.") {
    todo!("Clean up the whole check_migrate() function after v0.15 - it's not needed anymore");
}
```

The reminder is worth having. The mechanism means that once the version moves past 0.15, every debug and test build of every downstream consumer panics at node startup, since `check_migrate` is called unconditionally from `start_node_inner` under the `cache` feature. A `#[test]` asserting `APP_VERSION.starts_with("0.15.")`, or a `const _: () = assert!(...)`, delivers the same reminder to you at CI time without shipping a reachable panic into consumers.

Related, `:196-200`: the accepted-version allow-list is `0.15.`, `0.16.`, `1.0.`. A peer on `0.17.` would be rejected with "Invalid remote node version", which matters only if the module outlives the cleanup the `todo!` is about, but the two failure modes reinforce each other.

### 3. The snapshot publication fix is missing its fsync

`snapshot_builder.rs:51-60` now does the right thing structurally:

```rust
let path_tmp = format!("{path}~");
fs::copy(&path_temp, &path_tmp).await?;
if let Err(err) = fs::rename(&path_tmp, &path).await { ... }
```

That closes the non-atomic publication, which I had written up separately, so thank you for it. What it does not do is `sync_data()` the copied file before the rename. The rename is atomic with respect to the directory entry, but the copied bytes are not necessarily durable, so a crash between the copy and the writeback can leave a renamed but short snapshot that a restart will happily select.

Worth noting that the two cache-metadata paths added in this same PR do call `file.sync_data()` before their renames, so this reads as an omission rather than a decision.

### 4. `/version` is unauthenticated on the external API router

`api.rs`, `pub async fn get_version() -> impl IntoResponse { APP_VERSION }`, registered on both the internal router and the external one in `start.rs`. Anyone who can reach the API port learns the exact version.

Minor, but for a PR titled "Robustness and Security hardening" it stands out, and the only consumer that needs it, the migration check, already holds `secret_api` and could send the `X-API-SECRET` header like every other management call does.

### 5. A typo in the rename table

`CHANGELOG.md:63`:

```
health_check_delay          -> health_check_delay
```

The left side should be `health_check_delay_secs`.

### 6. Still open after this PR, in case any are in scope

These are things I had filed on the fork and checked against this PR; each is untouched here. Listing them purely so you know they are not covered, not as a request.

- `hiqlite/src/tls.rs` is untouched entirely. `ServerTlsConfig::from_env` falls through silently when only one of key/cert is set (plaintext, or an unverified self-signed cert if `HQL_TLS_AUTO_CERTS` is on), and `HQL_TLS_AUTO_CERTS` is `parse().unwrap_or(false)` while `HQL_TLS_{variant}_DANGER_TLS_NO_VERIFY` four lines later is `parse().expect(...)`, so the same typo is silent in one and fatal in the other.
- `config_toml.rs:211-212` reads `tls_raft_danger_tls_no_verify` a second time into `tls_api_danger_tls_no_verify`; since `t_bool` removes the key on first read, the API side is always `false` and the documented `tls_api_danger_tls_no_verify` key reaches the unknown-key check, refusing the whole config. Fail-closed, and I have executable tests for both halves.
- `hiqlite/src/store/logs/memory.rs` is untouched apart from import ordering, and has several divergences from the locked `openraft` 0.9.24 `RaftLogStorage` contract, including `purge` removing exclusively where the trait says inclusive.
- `log_io_completed` does not appear in this PR, so the WAL append completion contract is unchanged.
- With `tls_raft` and `tls_api` both set, every `tx_shutdown` watch receiver has been dropped by the time `shutdown_execute` sends, and the send is `expect`ed at `client/mgmt.rs`, so a graceful shutdown panics. The new concurrent-shutdown guard adds a `subscribe()` on the loser path only, so the single-caller case is unchanged.
- `build_listen_addr` splits the advertised address at the first `:`, so a bracketed IPv6 literal produces a malformed listen address. Your eager `SocketAddr::from_str` turns that from a silent missing listener into a loud startup error, which is a genuine improvement, but IPv6 node addresses still cannot be used.

I will open an issue with the full write-up so this PR does not become the place for it. Each of those has a traced line reference and, where relevant, a test.

bartekus:
One more thing on this PR, from a second pass over the same tree. It is not about a line this PR changes; it is about the same **class** of defect this PR fixes, in a file it does not touch.

### The snapshot publication fix has an exact counterpart in backup restore

`snapshot_builder.rs` now stages into `{path}~` and renames, which is what closes the non-atomic publication. `create_backup` (`store/state_machine/sqlite/writer.rs:790-806`) does the same thing, with a comment saying why.

`restore_backup` (`backup.rs:305-377`) does not:

```rust
// :351  validate first, which is the right order
is_metadata_ok(path_backup.clone()).await?;

// :354-358  then destroy
let _ = fs::remove_dir_all(&path_db).await;
let _ = fs::remove_dir_all(&path_snapshots).await;
let _ = fs::remove_dir_all(&path_lock_file).await;
let _ = fs::remove_dir_all(&path_logs).await;

fs::create_dir_all(&path_db).await?;
set_path_access(&path_db, 0o700).await?;

// :368  and only then put the replacement in place
fs::copy(&path_backup, &path_db_full).await?;
```

The validation happening before the destruction is right. The copy happening after it is the part that matches what this PR fixed elsewhere: a failure at `:368`, or of the `create_dir_all` or `set_path_access` between, leaves the node with neither its previous state nor the backup, and there is no rollback. The copied file is also never `sync_data`ed before the node starts on it, so a crash between the copy and the writeback can leave a short image that the next start accepts. That is the same fsync gap I mentioned for the snapshot path, one file over.

Staging the copy as `{path}~`, `sync_data`, then rename, and only removing the old directories after the rename succeeds, would make this match the shape the rest of the PR establishes.

### Two smaller ones in the same file, while I am here

**`is_metadata_ok` reports a corrupt backup as a task panic, not as invalid.** `backup.rs:391`:

```rust
let _meta: StateMachineData = deserialize(&bytes).unwrap();
```

This runs inside `spawn_blocking`, so a `_metadata` row that exists but does not decode panics the blocking task and the caller sees a join error rather than "this backup is not valid". The `TODO` two lines below already notes the validation is thinner than it looks.

**The post-restore log purge retries forever with no delay.** `backup.rs:477-479`:

```rust
while let Err(err) = state.raft_db.raft.trigger().purge_log(last_log).await {
    error!("Error during logs purge: {}", err);
}
```

No sleep, no bound, no exit. If the purge keeps failing, for example because the Raft is shutting down, this spins the task at full CPU emitting an error line per iteration. Every other loop in `restore_backup_finish` sleeps between 50 ms and 100 ms, and the snapshot trigger twelve lines above bails out rather than looping, which is the shape this one wants.

---

None of these is a request to widen this PR. They are recorded so the restore path is not left as the one place where the publication pattern this PR establishes does not apply. The rest of the second pass is going to #369; anything security-sensitive went through the private advisory channel instead.

sebadob:
This is still a Draft, so not finished yet. I do have quite a few issues already on my TODO, but not pushed them yet.
However, you had findings that would have slipped through probably, so this is very welcome, thanks.

> The v0.15 cache migration safeguard fails open

That is on purpose, and actually the only real "solution". It's **impossible** to guarantee correctness in this case, because there are so many factors we don't control. If you expect nodes to be online and on the new version upfront, this would be a state you can never get into. The API is only ever started after this check and cleanup. 

The thing about TLS: yes the client can simply always ignore validation for the checks, but then the whole old cluster would not be working anyway, which means it's impossible to sync old data from peers, which makes it somewhat "safe" again. And safe in this case does not mean bullet proof. This is impossible. It's meant to be an additional safety-net for people doing a major version upgrade of a database (we are below v1) without reading a changelog. As long as they had a working cluster before, this will catch it, and a non-working one will not sync bad data. They also only ever get to this step if they actually set the env var to trigger the automation, which means they either had the first errror and still did not read the changelog, or they read the changelog and just forgot a node (which is literally their only job to do).

> A read_dir failure for a reason other than "missing", a permissions problem for instance, leaves exit_early true, so the function concludes "No existing cache data found" and returns Ok. That is the same fail-open shape: an error about the data is read as an absence of data.

True, but if there is such an underlying error with the OS / path, the cache layer would not be able to start up either. I can add some additional checks though.

> 2. todo!() in library startup

100% on purpose. I want this to panic, and it's impossible to ever slip through into a release, because it will fire as soon as I bump it to the next version, whichever it will be. That's exactly why it's a `todo!()` and not just an error log, which would be missed very easily.

> 4. /version is unauthenticated on the external API router

I know, also kind of on purpose. I like having `/version` endpoints for my applications, which makes automated checks and audits very easy. Rauthy e.g. has one as well, so if you get a secrutiy warning for some version in the future, you can easily check all your deployments with a simple script. 

You could argue that it's basically free information for an attacker, and that's true (note only Raft-internal, but also the Raft API should never be public), but this would only be andvantage if you ignore important updates. You can catch these missing updates with automated auditing via `/version` endpoints though. An attacker can find out the running version in lots of ways usually, and the list is short, so it's not security sensitive information in my eyes. Having such an endpoint is rather a boost for your own, if you make use of it.

> build_listen_addr splits the advertised address at the first :, so a bracketed IPv6 literal produces a malformed listen address. Your eager SocketAddr::from_str turns that from a silent missing listener into a loud startup error, which is a genuine improvement, but IPv6 node addresses still cannot be used.

Would have missed that, since I never use IPv6, thanks.

---

I also saw the issues you opened. I won't reply on each of them now. I had a first look at them, and while I do have some of these findings on my TODO (or they are fixed already), I would definitely have missed others, so thank you!

The end goal is to make Hiqlite just absolutely stable. Until now, my main focus was on geting the automatic Raft and Cluster formation right, especially in the real world. There you have rolling restarts, lots of different timeouts, network issues, and so on. All of this is pretty stable in the latest version already, and my idea is that this PR will make it through the finish line (hopyfully). So, next thing is harden all the rest. Also, things like the split brain checker are probabably going away entirely. They were only added for better debugging in the initial phase.

